PR #105, opened 2015-01-30
CP-10878: Create planex-init command

ravippandey:


euanh:
Hi Ravi,

This looks like a good start, but there are a few more things to do.
- If the Makefile already exists, init should regenerate the part between the special planex-init comments but preserve any other contents of the file.   The code here just exits early if Makefile already exists.
- Please use the Python logging package instead of print_col.   See clone.py for an example.
- Please make sure that `pylint -rn init.py` and `pep8 init.py` print no warnings or errors.   Currently the code has the following classes of warnings:
  - indentation: use multiples of 4 spaces
  - whitespace: leave a blank space after commas in argument lists
  - docstrings: all functions need docstrings, and an overall docstring is needed for the file
  - constant names must be capitalized
  - single character variable names are not allowed
  - two blank lines are expected between functions

euanh:
Hi Ravi,

The code looks better but the algorithm is still wrong.   As it stands, you will leave the makefile alone if the first line matches the opening sentinel value; otherwise you will rewrite the entire makefile, appending the contents of the existing file after the planex-init stanza.   For example:

```
$ cat Makefile
some stuff here (possibly even a blank line)
# Start generated by planex-init
include /usr/share/planex/Makefile.rules
# End generated by planex-init
more stuff here

$ planex-init

$ cat Makefile
# Start generated by planex-init
include /usr/share/planex/Makefile.rules
more stuff added to planex-init since the last run
# End generated by planex-init
some stuff here (possibly even a blank line)
# Start generated by planex-init
include /usr/share/planex/Makefile.rules
# End generated by planex-init
more stuff here
```

What we actually want is the following:
```
$ cat Makefile
some stuff here (possibly even a blank line)
# Start generated by planex-init
include /usr/share/planex/Makefile.rules
# End generated by planex-init
more stuff here

$ planex-init

$ cat Makefile
some stuff here (possibly even a blank line)
# Start generated by planex-init
include /usr/share/planex/Makefile.rules
more stuff added to planex-init since the last run
# End generated by planex-init
more stuff here
```
This is pretty much how ```oasis setup``` works - you can take a look at that for more background.    I suggest that you code up a few test cases along these lines and use them to try out the code. 

In this case, you basically just need to loop over the existing Makefile, copying lines to the new one until you hit the opening planex marker.   Then put in the new auto-generated text, scan forward in the old Makefile until you reach the closing marker, then resume copying any further lines from the old Makefile into the new one.

planex/init.py
```diff
@@ -0,0 +1,49 @@
...
+    autogen = "include %s/Makefile.rules\n" % (MAKEFILE_PATH)
+    endline = "# End generated by planex-init\n"
+
+    if not os.path.exists(name):
+        log.debug("Creating Makefile")
+        with open(name, 'w') as h:
+            h.write(firstline)
+            h.write(autogen)
+            h.write(endline)
+
```
euanh: You can exit(0) here, to avoid going through the rewriting procedure when you have just created the file.

planex/init.py
```diff
@@ -0,0 +1,49 @@
...
+        lines = lines[:s+1] + [autogen] + lines[e:]
+    except:
+        raise
+
+    with open(name, 'w') as h:
+        h.writelines(lines)
+
+
+def main():
+    create_makefile()
```
euanh: Could you please add the following lines here, so this script can be run directly when testing (see configure.py for example)?
```
if __name__ == "__main__":
    main()
```

planex/init.py
```diff
@@ -0,0 +1,49 @@
...
+    with open(name, 'r') as h:
+        lines = h.readlines()
+
+    try:
+        s = lines.index(firstline)
+        e = lines.index(endline)
+
+        lines = lines[:s+1] + [autogen] + lines[e:]
+    except:
+        raise
```
euanh: Can you catch ValueError here and log an error with a friendlier message?

planex/init.py
```diff
@@ -0,0 +1,55 @@
...
+    with open(name, 'r') as h:
+        lines = h.readlines()
+
+    try:
+        s = lines.index(firstline)
+        e = lines.index(endline)
+
+        lines = lines[:s+1] + [autogen] + lines[e:]
+    except ValueError:
+        log.setLevel(logging.ERROR)
```
euanh: I don't think you need this line.

planex/init.py
```diff
@@ -0,0 +1,55 @@
...
+Creates or regenerates a Makefile with special planex-init comments
+"""
+
+import os
+import logging
+
+MAKEFILE_PATH = "/usr/share/planex"
+
+log = logging.getLogger(__name__)
+logging.basicConfig(level=logging.DEBUG)
```
euanh: Let's default to logging.ERROR.

PR #135, opened 2015-03-20
fetch: Download all sources defined in the spec file

### LINKED ISSUE #130: planex-fetch cannot download multiple sources for the same spec file
planex-fetch currently takes one spec file name and one source file name as arguments, and downloads the source file as defined in the spec.  If a spec file defines multiple sources, multiple invocations of planex-fetch are needed.   However planex-fetch is usually run by a make pattern rule.   The dependency rules generated by planex-depend to download sources are of the form:

    foo.tar.gz: foo.spec
    foo-bar.tar.gz: foo.spec

Make condenses these into:

    foo.tar.gz foo-bar.tar.gz: foo.spec

make only passes one of the targets to the planex-fetch invocation [1], so the other target is not downloaded.   Furthermore, because the first target is downloaded, the pattern rule will never fire again.

[]1] See https://www.gnu.org/software/make/manual/html_node/Pattern-Intro.html#Pattern-Intro for details.

The solution is either to make planex-fetch able to download several sources files in one invocation, then find a way to get make to pass all the target names to it, or to add an option to planex-fetch which makes it download all sources defined in the spec file.

euanh:
The make pattern rule which calls fetch only fires once, even if the
rule defines multiple targets.   Make does not provide a variable to
access all the targets, only the one which happened to be the first to
need to be rebuilt.   To handle a spec file which needs several sources
we must download all of them in one invocation of fetch.

Fixes #130.

### DIFF
--- a/planex/fetch.py
+++ b/planex/fetch.py
@@ -73,18 +73,13 @@ def fetch_http(url, filename, retries):
                 raise
 
 
-def url_for_source(spec, source):
+def all_sources(spec):
     """
-    Find the URL corresponding to source in the spec file
+    Get all source URLs defined in the spec file
     """
-    spec = planex.spec.Spec(spec)
-    source_basename = os.path.basename(source)
-
-    for path, url in zip(spec.source_paths(), spec.source_urls()):
-        if path.endswith(source_basename):
-            return urlparse.urlparse(url)
-
-    raise KeyError(source_basename)
+    spec = planex.spec.Spec(spec, topdir=".")
+    urls = [urlparse.urlparse(url) for url in spec.source_urls()]
+    return zip(spec.source_paths(), urls)
 
 
 def parse_args_or_exit(argv=None):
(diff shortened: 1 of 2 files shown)


euanh:
@simonjbeaumont: Could you take a moment to review this please?

planex/fetch.py
```diff
@@ -73,18 +73,13 @@ def fetch_http(url, filename, retries):
...
     """
-    spec = planex.spec.Spec(spec)
-    source_basename = os.path.basename(source)
-
-    for path, url in zip(spec.source_paths(), spec.source_urls()):
-        if path.endswith(source_basename):
-            return urlparse.urlparse(url)
-
-    raise KeyError(source_basename)
+    spec = planex.spec.Spec(spec, topdir=".")
```
simonjbeaumont: `topdir` wasn't specified before this?

simonjbeaumont:
All sane except for the hard-coding of `topdir` (commented inline).

euanh:
@simonjbeaumont: I've added a --topdir flag to fetch.

PR #137, opened 2015-03-24
Further pylint improvements

euanh:
Fix a few more problems identified by pylint and tighten the scan after removing planex-configure et al.


simonjbeaumont:
:shipit:

PR #139, opened 2015-03-24
Quieten cache

euanh:


euanh:
@simonjbeaumont: This should make cache less chatty.

PR #148, opened 2015-03-26
Add bash completion for all entry points

### LINKED ISSUE #134: Add bash completion support
http://tldp.org/LDP/abs/html/tabexpansion.html

simonjbeaumont:
Fixes #134.

### DIFF
--- a/planex.spec
+++ b/planex.spec
@@ -15,6 +15,7 @@ Requires: mock
 Requires: rpm-build
 Requires: createrepo
 Requires: python-argparse
+Requires: python-argcomplete
 Requires: python-setuptools
 
 %description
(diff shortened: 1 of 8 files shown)


simonjbeaumont:
Query packaging... There's a python module dependency we need to install. I don't think it can be added to the spec file cos it's not in the normal yum repos. It is available in `pip` and I have added it to `requirements.txt` so it will be installed by Travis. However, will this be automagically packaged up as part of setuptools?

simonjbeaumont:
Huh. It is there in the yum repo... I'll add it to the spec file.

simonjbeaumont:
Once this is merged, all the necessary stuff will be in `/etc/bash_completion.d/`. All a user would need to do then is install `bash-completion` on their distro.

euanh:
Looks good, thanks!

Issue #146, opened 2015-03-25
Makefile.rules should use variables to represent all commands and options


euanh:
Make's implicit rules use variables to represent commands and their options, e.g. $(CC) $(CFLAGS).  Doing this in Makefile.rules would make it easy to override a particular command or add options.

PR #155, opened 2015-03-26
Makefile: Define variables for major commands and their arguments

euanh:


euanh:
Fixes #146

PR #165, opened 2015-03-31
pin: Delete dict of spec pins when all its source pins are removed

### LINKED ISSUE #164: planex-pin remove should not leave an empty dict behind
```
planex-pin add SPECS/planex.spec  ../planex#master
planex-pin remove SPECS/planex.spec
 cat pins
# This file is auto-generated by planex-pin
# Do not edit directly, instead use planex-pin {add,list,remove}
{"SPECS/planex.spec": {}}
```

`pins` should be empty when all pins have been removed

simonjbeaumont:
Fixes #164 


euanh:
:shipit:

PR #170, opened 2015-03-31
fetch: Use a temporary file to avoid blatting on error

simonjbeaumont:


johnelse:
:trollface:

Issue #150, opened 2015-03-26
Concurrent createrepo runs can race and cause repo metadata corruption


euanh:
Probably caused by non-atomic updates of repo metadata.

```
Traceback (most recent call last):
File "/usr/bin/planex-cache", line 9, in <module>
load_entry_point('planex==0.6.0', 'console_scripts', 'planex-cache')()
File "/usr/lib/python2.7/site-packages/planex-0.6.0-py2.7.egg/planex/cache.py", line 239, in
_main
main(sys.argv[1:])
File "/usr/lib/python2.7/site-packages/planex-0.6.0-py2.7.egg/planex/cache.py", line 217, in
main
pkg_hash = get_srpm_hash(srpm, yumbase, mock_config)
File "/usr/lib/python2.7/site-packages/planex-0.6.0-py2.7.egg/planex/cache.py", line 156, in
get_srpm_hash
pkgs = yumbase.pkgSack.returnNewestByNameArch(patterns=[req])
File "/usr/lib/python2.7/site-packages/yum/packageSack.py", line 520, in
returnNewestByNameArch
pkgs = calr("returnNewestByNameArch", naTup, patterns, ignore_case)
File "/usr/lib/python2.7/site-packages/yum/packageSack.py", line 595, in
_computeAggregateListResult
sackResult = apply(method, args)
File "/usr/lib/python2.7/site-packages/yum/sqlitesack.py", line 46, in newFunc
return func(*args, **kwargs)
File "/usr/lib/python2.7/site-packages/yum/sqlitesack.py", line 1472, in
returnNewestByNameArch
ignore_case)
File "/usr/lib/python2.7/site-packages/yum/packageSack.py", line 951, in
returnNewestByNameArch
ignore_case=ignore_case)
File "/usr/lib/python2.7/site-packages/yum/sqlitesack.py", line 1669, in returnPackages
pkgobjlist = self._buildPkgObjList(repoid, patterns, ignore_case)
File "/usr/lib/python2.7/site-packages/yum/sqlitesack.py", line 46, in newFunc
return func(*args, **kwargs)
File "/usr/lib/python2.7/site-packages/yum/sqlitesack.py", line 1628, in _buildPkgObjList
return self.searchNames(patterns)
File "/usr/lib/python2.7/site-packages/yum/sqlitesack.py", line 57, in newFunc
raise Errors.RepoError, str(e)
yum.Errors.RepoError: database disk image is malformed
[...]

PR #180, opened 2015-04-01
Makefile: Improve locking around repo metadata

euanh:
There can be a race between mock consuming repository
metadata and createrepo producing it.   This is especially
apparent when planex-cache is used instead of mock - it also
consumes the repository metadata, but cycles so quickly that
the likelihood of a race is much greater than with conventional
mock.

This changeset wraps mock/planex-cache with a shared (a.k.a. read)
lock on the RPMS directory.   The createrepo step already takes an
exclusive (a.k.a. write) lock on that directory - this changeset
makes that more explicit by passing --exclusive, which is the default
for flock.

This change should prevent races between mock and createrepo, but
it may come at the cost of some performance.   An atomic
compare-and-swap wrapper around createrepo might be better, but
the interaction with yum would still be tricky.

### DIFF
--- a/planex/Makefile.rules
+++ b/planex/Makefile.rules
@@ -104,11 +104,16 @@ srpm_repo: srpms
 # The repository metadata is updated after building a binary package so that
 # a subsequent mock build for a package which depend on this one is able
 # to find and install it.
+#
+# We take a shared lock on RPMS around mock to prevent concurrent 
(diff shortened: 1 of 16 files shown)


euanh:
Fixes #150

euanh:
This doesn't seem to fix the problem yet.

PR #193, opened 2015-04-07
fetch: Check downloaded content looks sensible

### LINKED ISSUE #123: Add file type checking to planex-fetch
planex-downloader checked whether the result of a download looked like an archive, and retried the download if it didn't.  planex-fetch should do the same thing.

There is a Python library to wrap around libmagic, but it is more or less as ugly as just shelling out to file, as downloader used to do.

simonjbeaumont:
Fixes #123


euanh:
:shipit:

PR #198, opened 2015-04-17
Fetch from mirror

jonludlam:
Plus Euan's PR


planex/cache.py
```diff
@@ -33,6 +33,9 @@ def parse_args_or_exit(argv=None):
         '--loopback-repo', default='mock',
         help='Name of the mock loopback repository')
     parser.add_argument(
```
euanh: We should not need this.   Set the PATH environment variable correctly, with /usr/bin at the front, so /usr/bin/mock is found instead of /usr/sbin/mock.

PR #205, opened 2015-04-23
Set up a better logging format...

simonjbeaumont:
Now a log line will look like the following:

    Apr 23 10:46:09.225 planex-pin [30369] debug: Configured logging.

with the PID in the square brackets.


simonjbeaumont:
Fixes #202

PR #214, opened 2015-04-24
Default to MD5 if srpm.filedigestalgo isn't set

johnelse:


thomassa:
Why MD5?

johnelse:
This code is only to log the algorithm that's already been used. Older versions of the library don't set the `filedigestalgo` field, but we found MD5 had been used so was a sensible default.

euanh:
MD5 was the default digest algorithm in RPM 4.4 and earlier.  In RPM 4.6 the digest algorithm became configurable: http://rpm.org/wiki/releases/4.6

thomassa:
Makes sense then.

PR #219, opened 2015-05-20
cache: Print mock's logs if it fails

euanh:
We print mock's stderr and stdout if it fails, but the most useful
information for debugging a mock failure is often in the root.log
and build.log files.    Print these as well.


jonludlam:
Looks good to me!

PR #220, opened 2015-05-21
cache: Update cached files' timestamps on cache hits

euanh:
Update a cached file's access and modification timestamps when we use it.
This makes it easier to clear old, unused files out of the cache.


jonludlam:
you should probably wrap the utime thing in an exception handler to cover the case of a RO cache dir. This bit us last time :-)

PR #225, opened 2015-08-18
Depend on yum-plugin-priorities

euanh:
We set the priority of the loopback repository such that it overrides the base system repository.  Without this plugin, priority assignments in yum.conf.d are ignored.


euanh:
Fixes #233

PR #239, opened 2015-11-10
planex-cache: Don't fail if the binary package is already in the cache

### LINKED ISSUE #238: planex-cache should not assert and exit if binary RPM is already in the cache when it comes to write back
planex-cache can fail if another racing build writes a binary RPM to the cache between the time when cache checks the cache and tries to write back to it after building the package.   The race is here:
https://github.com/xenserver/planex/blob/a9dbb0b36095625df5eebf611391e4692db6d8e8/planex/cache.py#L235-L241

add_to_cache() asserts that the cache directory doesn't already exist, which is fine during development on a single machine, but wrong for a shared cache in a busy build system:
https://github.com/xenserver/planex/blob/a9dbb0b36095625df5eebf611391e4692db6d8e8/planex/cache.py#L96

If we trust our hashes, we can just drop this assert and short-circuit add_to_cache() if the directory already exists. If we are more paranoid, we can check whether the cached RPM looks plausible. It shouldn't be a hard fix either way.

euanh:
planex-cache can fail if another racing build writes a binary RPM to the
cache between the time when cache checks the cache and tries to write
back to it after building the package.   If the package is already in the
cache, we can just throw away our local build.

Fixes #238.

### DIFF
--- a/planex/cache.py
+++ b/planex/cache.py
@@ -92,9 +92,15 @@ def add_to_cache(cachedirs, pkg_hash, build_dir):
     """
     Add the build products in build_dir to the cache
     """
-    cache_dir = cache_locations(cachedirs, pkg_hash)[0]
-    assert not os.path.isdir(cache_dir)
 
+    # Another racing build may have written this package into the cache
+    # while we were building.  If that is the case, we don't have to do
+    # anything.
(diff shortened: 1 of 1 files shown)


jonludlam:
This looks good to me.

PR #259, opened 2016-03-15
cache: Make sure we call /usr/bin/mock instead of /usr/sbin/mock

euanh:
Mock should be run by a normal user account calling /usr/bin/mock.
If the user's path causes /usr/sbin/mock to be picked up instead,
it will exit with a warning but with a 0 exit code.
Subprocess.Popen will not detect this as an error and the build will
appear to succeed without building anything.


jonludlam:
We can remove the path hack in Makefile.xs now.

:+1:

PR #260, opened 2016-03-18
makesrpm: create a wrapper around rpmbuild

salvocambria:
Signed-off-by: Salvatore Cambria <salvatore.cambria@citrix.com>

### DIFF
--- a/planex/makesrpm.py
+++ b/planex/makesrpm.py
@@ -27,7 +27,7 @@ def parse_args_or_exit(argv):
         "--topdir", metavar="DIR", default=None,
         help='Set rpmbuild toplevel directory')
     parser.add_argument(
-        "--dist", metavar="DIST", default="",
+        "--dist", metavar="DIST", default=None,
         help="distribution tag (used in RPM filenames)")
     argcomplete.autocomplete(parser)
     return parser.parse_known_args(argv)
@@ -38,22 +38,23 @@ def parse_patchseries(series_file):
     Parse series file and return the list of patches
     """
     patches = []
-    for line in open(series_file):
-        line = line.strip()
-        if not line.startswith("#"):
-            patches.append(line)
+    with open(series_file) as series:
+        for line in series:
+            line = line.partition('#')[0].strip()
+            if line:
+                patches.append(line)
 
     return patches
 
 
 def setup_tmp_area():
     """
-    Create temporary wotking area
+    Create temporary working area
     """
-    tmp_dirpath = tempfile.mkdtemp() + '/'
-    tmp_specs = tmp_dirpath + 'SPECS/'
-    tmp_build = tmp_dirpath + '_build/'
-    tmp_sources = tmp_build + 'SOURCES/'
+    tmp_dirpath = tempfile.mkdtemp()
+    tmp_specs = os.path.join(tmp_dirpath, 'SPECS')
+    tmp_build = os.path.join(tmp_dirpath, '_build')
+    tmp_sources = os.path.join(tmp_build, 'SOURCES')
 
     os.makedirs(tmp_specs)
     os.makedirs(tmp_build)
@@ -62,72 +63,83 @@ def setup_tmp_area():
     return (tmp_dirpath, tmp_sources)
 
 
+def extract_patches(tmp_specfile, patchqueue_filters, patchqueue_path,
+                    tmp_sources, target):
+    """
+    Extract patches and inject them into the specfile
+    """
+    for line in fileinput.input(tmp_specfile, inplace=True):
+        if any([ext in line for ext in patchqueue_filters]):
+            tar = tarfile.open(patchqueue_path)
+            for mem in tar.getmembers():
+                # Modify mem.name in place to change the name of the
+                # extracted file (and drop any leading paths)
+                mem.name = '%s-' % target + os.path.basename(mem.name)
+                tar.extract(mem, tmp_sources)
+            patches = parse_patchseries(os.path.join(tmp_sources,
+                                                     '%s-series' % target))
+            print "# Patches for %s" % target
+            for patch_num, patch in enumerate(patches):
+                if patch:
+                    print "Patch%s: %%{name}-%s" % (patch_num, patch)
+        else:
+            print line,
+
+
+def get_command_line(intercepted_args, tmp_sources, tmp_specfile):
+    """
+    Return rpmbuild command line and arguments
+    """
+    cmd = ['rpmbuild']
+    if intercepted_args.quiet:
+        cmd.append('--quiet')
+    if intercepted_args.topdir is not None:
+        cmd.append('--define')
+        cmd.append('_topdir %s' % intercepted_args.topdir)
+    if intercepted_args.dist is not None:
+        cmd.append('--define')
+        cmd.append('%%dist %s' % intercepted_args.dist)
+    cmd.append('--define')
+    cmd.append('_sourcedir %s' % tmp_sources)
+    cmd.append('-bs')
+    cmd.append(tmp_specfile)
+
+    return cmd
+
+
 def main(argv):
     """
     Entry point arguments: component.spec Source0 Source1 ... Patch0 Patch1 ...
     The specfile must always be present and it has to be the first in the list
     """
     intercepted_args, passthrough_args = parse_args_or_exit(argv)
-    target = os.path.basename(passthrough_args[0]).split('.')[0]
+    target = os.path.splitext(os.path.basename(passthrough_args[0]))[0]
     specfile = passthrough_args[0]
     tmp_dirpath, tmp_sources = setup_tmp_area()
-    tmp_specfile = tmp_dirpath + specfile
+    tmp_specfile = os.path.join(tmp_dirpath, specfile)
 
     try:
         # Copy files to temporary working area
         copyfile(specfile, tmp_specfile)
-        patchqueue = False
-        patchqueue_path = None
         patchqueue_filters = ['.pg.', '.pq.']
 
         for source in passthrough_args[1:]:
-            copyfile(source, tmp_dirpath + source)
             if any([ext in source for ext in patchqueue_filters]):
-                patchqueue = True
-                patchqueue_path = source
+                extract_patches(tmp_specfile, patchqueue_filters,
+                                source, tmp_sources, target)
+            else:
+                copyfile(source, os.path.join(tmp_dirpath, source))
 
-        if patchqueue:
-            # Extract patches
-            patch_num = 0
-            for line in fileinput.input(tmp_specfile, inplace=True):
-                if any([ext in line for ext in patchqueue_filters]):
-                    tar = tarfile.open(tmp_dirpath + patchqueue_path)
-                    for mem in tar.getmembers():
-                        mem.name = '%s-' % target + os.path.basename(mem.name)
-                        tar.extract(mem, tmp_sources)
-                    patches = parse_patchseries(tmp_sources +
-                                                '%s-series' % target)
-                    print "# Patches for %s" % target
-                    for patch in patches:
-                        if patch != "":
-                            print "Patch%s: %%{name}-%s" % (patch_num, patch)
-                            patch_num += 1
-                else:
-                    print line,
-
-        args = []
-        if intercepted_args.quiet:
-            args.append('--quiet')
-        if intercepted_args.topdir is not None:
-            args.append('--define')
-            args.append('_topdir %s' % intercepted_args.topdir)
-        if intercepted_args.dist != "":
-            args.append('--define')
-            args.append(u'%dist ' + intercepted_args.dist)
-        args.append('--define')
-        args.append('_sourcedir %s' % tmp_sources)
-        args.append('-bs')
-        args.append(tmp_specfile)
-        cmd = ['rpmbuild']
-        cmd.extend(args)
-
+        cmd = get_command_line(intercepted_args, tmp_sources, tmp_specfile)
         return_value = subprocess.call(cmd)
         sys.exit(return_value)
 
-    except Exception as inst:
-        print type(inst)
-        print inst.args
-        print inst
+    except IOError as exc:
+        print "Copyfile: destination location must be writable"
+        print "Exception: %s" % exc
+    except (tarfile.TarError, tarfile.ReadError) as exc:
+        print "Error when extracting patchqueue from tarfile"
+        print "Exception: %s" % exc
 
     finally:
         # Clean temporary area
--- a/planex/makesrpm.py
+++ b/planex/makesrpm.py
@@ -1,0 +1,145 @@
+#!/usr/bin/python
+
+"""
+planex-make-srpm: Wrapper around rpmbuild
+"""
+
+import sys
+import subprocess
+import os
+from shutil import copyfile, rmtree
+import fileinput
+import tarfile
+import tempfile
+
+import argparse
+import argcomplete
+from planex.util import add_common_parser_options
+
+
+def parse_args_or_exit(argv):
+    """
+    Parse command line options
+    """
+    parser = argparse.ArgumentParser(description='Planex SRPM building')
+    add_common_parser_options(parser)
+    parser.add_argument(
+        "--topdir", metavar="DIR", default=None,
+        help='Set rpmbuild toplevel directory')
+    parser.add_argument(
+        "--dist", metavar="DIST", default="",
+        help="distribution tag (used in RPM filenames)")
+    argcomplete.autocomplete(parser)
+    return parser.parse_known_args(argv)
+
+
+def parse_patchseries(series_file):
+    """
+    Parse series file and return the list of patches
+    """
+    patches = []
+    for line in open(series_file):
+        line = line.strip()
+        if not line.startswith("#"):
+            patches.append(line)
+
+    return patches
+
+
+def setup_tmp_area():
+    """
+    Create temporary wotking area
+    """
+    tmp_dirpath = tempfile.mkdtemp() + '/'
+    tmp_specs = tmp_dirpath + 'SPECS/'
+    tmp_build = tmp_dirpath + '_build/'
+    tmp_sources = tmp_build + 'SOURCES/'
+
+    os.makedirs(tmp_specs)
+    os.makedirs(tmp_build)
+    os.makedirs(tmp_sources)
+
+    return (tmp_dirpath, tmp_sources)
+
+
+def main(argv):
+    """
+    Entry point arguments: component.spec Source0 Source1 ... Patch0 Patch1 ...
+    The specfile must always be present and it has to be the first in the list
+    """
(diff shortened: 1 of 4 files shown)


planex/makesrpm.py
```diff
@@ -0,0 +1,133 @@
...
+                    tar = tarfile.open(tmp_dirpath + patchqueue_path)
+                    for f in tar.getmembers():
+                        f.name = '%s-' % target + os.path.basename(f.name)
+                        tar.extract(f, tmp_sources)
+                    patches = parse_patchseries(tmp_sources +
+                                                '%s-series' % target)
+                    print "# Patches for %s" % target
+                    for p in patches:
+                        if p != "":
+                            print "Patch%s: " % patch_number +
```
euanh: You can do this in one string interpolation step:
`print "Patch%s: %%{name}-%s" % (patch_number, p)`

planex/makesrpm.py
```diff
@@ -0,0 +1,132 @@
...
+    tmp_build = tmp_dirpath + '_build/'
+    tmp_sources = tmp_build + 'SOURCES/'
+
+    # Create temporary working area
+    os.makedirs(tmp_specs)
+    os.makedirs(tmp_build)
+    os.makedirs(tmp_sources)
+
+    try:
+        # Copy files to temporary working area
```
euanh: I think it would be good to the logic that sets up the temporary working area into a separate function.

planex/makesrpm.py
```diff
@@ -0,0 +1,132 @@
...
+        copyfile(specfile, tmp_specfile)
+        patchqueue = False
+        patchqueue_path = None
+        patchqueue_filters = ['.pg.', '.pq.']
+
+        for source in passthrough_args[1:]:
+            copyfile(source, tmp_dirpath + source)
+            if any(map(lambda x: x in source, patchqueue_filters)):
```
euanh: this should be: `[ext in source for ext in patchqueue_filters]`

planex/makesrpm.py
```diff
@@ -0,0 +1,132 @@
...
+                patchqueue = True
+                patchqueue_path = source
+
+        if patchqueue:
+            # Extract patches
+            patch_number = 0
+            for line in fileinput.input(tmp_specfile, inplace=True):
+                if any(map(lambda x: x in line, patchqueue_filters)):
```
euanh: Same as before.

planex/makesrpm.py
```diff
@@ -0,0 +1,132 @@
+import sys
+import subprocess
+import os
+from shutil import copyfile
```
euanh: import copyfile, rmtree, and remove the lower import shutil.

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+        line = line.strip()
+        if not line.startswith("#"):
+            patches.append(line)
+
+    return patches
+
+
+def setup_tmp_area():
+    """
+    Create temporary wotking area
```
euanh: wotking -> working

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    tmp_dirpath = tempfile.mkdtemp() + '/'
+    tmp_specs = tmp_dirpath + 'SPECS/'
```
euanh: Please use os.path.join(tmp_dirpath, "SPECS")
If you do that, you also don't need to append '/' to tmp_dirpath

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+                                                '%s-series' % target)
+                    print "# Patches for %s" % target
+                    for patch in patches:
+                        if patch != "":
+                            print "Patch%s: %%{name}-%s" % (patch_num, patch)
+                            patch_num += 1
+                else:
+                    print line,
+
+        args = []
```
euanh: Why not start with `args = ["rpmbuild"]`, then you can drop the cmd.extend(args) later on.

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+        if intercepted_args.dist != "":
+            args.append('--define')
+            args.append(u'%dist ' + intercepted_args.dist)
+        args.append('--define')
+        args.append('_sourcedir %s' % tmp_sources)
+        args.append('-bs')
+        args.append(tmp_specfile)
+        cmd = ['rpmbuild']
+        cmd.extend(args)
+
```
euanh: Drop `cmd = ... cmd.extend(...)`

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+        return_value = subprocess.call(cmd)
+        sys.exit(return_value)
+
+    except Exception as inst:
```
euanh: I'm afraid you need to figure out what specific exceptions you want to handle here.   The docs are usually pretty good at telling you what a function might throw.    Splitting up `main` into a few pieces should also help (as well as getting rid of the 'too many nested blocks' and 'too many statements' warnings.

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+
+    try:
+        # Copy files to temporary working area
+        copyfile(specfile, tmp_specfile)
+        patchqueue = False
+        patchqueue_path = None
+        patchqueue_filters = ['.pg.', '.pq.']
+
+        for source in passthrough_args[1:]:
+            copyfile(source, tmp_dirpath + source)
```
euanh: use `os.path.join` instead of concatenating `tmp_dirpath + source`

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+def main(argv):
+    """
+    Entry point arguments: component.spec Source0 Source1 ... Patch0 Patch1 ...
+    The specfile must always be present and it has to be the first in the list
+    """
+    intercepted_args, passthrough_args = parse_args_or_exit(argv)
+    target = os.path.basename(passthrough_args[0]).split('.')[0]
+    specfile = passthrough_args[0]
+    tmp_dirpath, tmp_sources = setup_tmp_area()
+    tmp_specfile = tmp_dirpath + specfile
```
euanh: use `os.path.join`

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+    return (tmp_dirpath, tmp_sources)
+
+
+def main(argv):
+    """
+    Entry point arguments: component.spec Source0 Source1 ... Patch0 Patch1 ...
+    The specfile must always be present and it has to be the first in the list
+    """
+    intercepted_args, passthrough_args = parse_args_or_exit(argv)
+    target = os.path.basename(passthrough_args[0]).split('.')[0]
```
euanh: You can use `os.path.splitext()` instead of `split('.')`

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+            if any([ext in source for ext in patchqueue_filters]):
+                patchqueue = True
+                patchqueue_path = source
+
+        if patchqueue:
+            # Extract patches
+            patch_num = 0
+            for line in fileinput.input(tmp_specfile, inplace=True):
+                if any([ext in line for ext in patchqueue_filters]):
+                    tar = tarfile.open(tmp_dirpath + patchqueue_path)
```
euanh: os.path.join

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def parse_patchseries(series_file):
+    """
+    Parse series file and return the list of patches
+    """
+    patches = []
+    for line in open(series_file):
+        line = line.strip()
+        if not line.startswith("#"):
```
euanh: You should also remove end-of-line comments here.   Something like this should work:
```
line = line.partition('#')[0].strip()
if line:
   patches.append(line)

File: planex/makesrpm.py [hunk 16]
```diff
@@ -0,0 +1,145 @@
...
+    argcomplete.autocomplete(parser)
+    return parser.parse_known_args(argv)
+
+
+def parse_patchseries(series_file):
+    """
+    Parse series file and return the list of patches
+    """
+    patches = []
+    for line in open(series_file):
```
euanh: It's better to use a context manager to close the file at the end of the block, rather than relying on the GC:
```
with open(series_file) as series:
   for line in series:
      ...

File: planex/makesrpm.py [hunk 17]
```diff
@@ -0,0 +1,145 @@
...
+                patchqueue_path = source
+
+        if patchqueue:
+            # Extract patches
+            patch_num = 0
+            for line in fileinput.input(tmp_specfile, inplace=True):
+                if any([ext in line for ext in patchqueue_filters]):
+                    tar = tarfile.open(tmp_dirpath + patchqueue_path)
+                    for mem in tar.getmembers():
+                        mem.name = '%s-' % target + os.path.basename(mem.name)
```
euanh: mem.name doesn't seem to be used anywhere

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+                patchqueue = True
+                patchqueue_path = source
+
+        if patchqueue:
+            # Extract patches
+            patch_num = 0
+            for line in fileinput.input(tmp_specfile, inplace=True):
+                if any([ext in line for ext in patchqueue_filters]):
+                    tar = tarfile.open(tmp_dirpath + patchqueue_path)
+                    for mem in tar.getmembers():
```
euanh: isn't this loop just the same as `tar.extractall(path=tmp_sources)`

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+    try:
+        # Copy files to temporary working area
+        copyfile(specfile, tmp_specfile)
+        patchqueue = False
+        patchqueue_path = None
+        patchqueue_filters = ['.pg.', '.pq.']
+
+        for source in passthrough_args[1:]:
+            copyfile(source, tmp_dirpath + source)
+            if any([ext in source for ext in patchqueue_filters]):
```
euanh: I think you should put this test at the top of the loop and decide whether to extract a patchqueue or just copy the source file at that point.

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+                patchqueue = True
+                patchqueue_path = source
+
+        if patchqueue:
+            # Extract patches
+            patch_num = 0
+            for line in fileinput.input(tmp_specfile, inplace=True):
```
euanh: Extract this loop into a separate function.

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+                    print "# Patches for %s" % target
+                    for patch in patches:
+                        if patch != "":
+                            print "Patch%s: %%{name}-%s" % (patch_num, patch)
+                            patch_num += 1
+                else:
+                    print line,
+
+        args = []
+        if intercepted_args.quiet:
```
euanh: It would be neater to pull this command-line setup into a separate function.

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+            args.append('--quiet')
+        if intercepted_args.topdir is not None:
+            args.append('--define')
+            args.append('_topdir %s' % intercepted_args.topdir)
+        if intercepted_args.dist != "":
+            args.append('--define')
+            args.append(u'%dist ' + intercepted_args.dist)
```
euanh: This line can be `args.append("%%dist %s" % intercepted_args.dist)`

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+            for line in fileinput.input(tmp_specfile, inplace=True):
+                if any([ext in line for ext in patchqueue_filters]):
+                    tar = tarfile.open(tmp_dirpath + patchqueue_path)
+                    for mem in tar.getmembers():
+                        mem.name = '%s-' % target + os.path.basename(mem.name)
+                        tar.extract(mem, tmp_sources)
+                    patches = parse_patchseries(tmp_sources +
+                                                '%s-series' % target)
+                    print "# Patches for %s" % target
+                    for patch in patches:
```
euanh: You can replace this with `for patch_num, patch in enumerate(patches)` and then drop the patch_num declaration above.

planex/makesrpm.py
```diff
@@ -0,0 +1,145 @@
...
+
+        for source in passthrough_args[1:]:
+            copyfile(source, tmp_dirpath + source)
+            if any([ext in source for ext in patchqueue_filters]):
+                patchqueue = True
+                patchqueue_path = source
+
+        if patchqueue:
+            # Extract patches
+            patch_num = 0
```
euanh: You don't need this if you use `enumerate(...)` below.

planex/makesrpm.py
salvocambria: I needed to keep the '/' when using os.path.join()

euanh: You don't need to keep the '/' if you use os.path.join everywhere you need to build a path.

salvocambria: Yes correct, I realised that it was failing for a different reason.

planex/makesrpm.py
salvocambria: I am modifying, in place, the name of each member of the tarball attaching the component's name at the very beginning. 
When extracting the member from the tarball with tar.extract(mem, tmp_source), it will look like '<target>-member'.

planex/makesrpm.py
salvocambria: It is needed in order to modify the name of each member before extracting it. We want each member of the tarball to look like target-member_name

planex/makesrpm.py
euanh: Hmm,  I don't like that very much.   I don't see anything in the tarfile docs which suggests that's a valid thing to do.   I think it would be clearer and safer to extract the file with its original name, then rename it.

planex/makesrpm.py
```diff
@@ -0,0 +1,160 @@
...
+
+        if patchqueue:
+            extract_patches(tmp_specfile, patchqueue_filters, patchqueue_path,
+                            tmp_sources, target)
+
+        cmd = get_command_line(intercepted_args, tmp_sources, tmp_specfile)
+        return_value = subprocess.call(cmd)
+        sys.exit(return_value)
+
+    except Exception as inst:
```
euanh: some exceptions you might want to consider handling here:
* Error and IOError from copyfile
* OSError and ValueError from subprocess.call (maybe)
* TarError and ReadError from tarfile.extract

planex/makesrpm.py
```diff
@@ -0,0 +1,160 @@
...
+    try:
+        # Copy files to temporary working area
+        copyfile(specfile, tmp_specfile)
+        patchqueue = False
+        patchqueue_path = None
+        patchqueue_filters = ['.pg.', '.pq.']
+
+        for source in passthrough_args[1:]:
+            if any([ext in source for ext in patchqueue_filters]):
+                patchqueue = True
```
euanh: I think you can just call extract_patches here.

planex/makesrpm.py
```diff
@@ -0,0 +1,160 @@
...
+                patchqueue_path = source
+            else:
+                copyfile(source, os.path.join(tmp_dirpath, source))
+
+        if patchqueue:
```
euanh: If we move extract_patches into the loop, we don't need to call it here.

planex/makesrpm.py
```diff
@@ -0,0 +1,160 @@
...
+    intercepted_args, passthrough_args = parse_args_or_exit(argv)
+    target = os.path.splitext(os.path.basename(passthrough_args[0]))[0]
+    specfile = passthrough_args[0]
+    tmp_dirpath, tmp_sources = setup_tmp_area()
+    tmp_specfile = os.path.join(tmp_dirpath, specfile)
+
+    try:
+        # Copy files to temporary working area
+        copyfile(specfile, tmp_specfile)
+        patchqueue = False
```
euanh: You can get rid of these variables if you move extract_patches as described below.

planex/makesrpm.py
```diff
@@ -0,0 +1,160 @@
...
+            for mem in tar.getmembers():
+                mem.name = '%s-' % target + os.path.basename(mem.name)
+                tar.extract(mem, tmp_sources)
+            patches = parse_patchseries(os.path.join(tmp_sources,
+                                                     '%s-series' % target))
+            print "# Patches for %s" % target
+            for patch_num, patch in enumerate(patches):
+                if patch != "":
+                    print "Patch%s: %%{name}-%s" % (patch_num, patch)
+                    patch_num += 1
```
euanh: You don't have to increment `patch_num` here - enumerate does that for you.

planex/makesrpm.py
```diff
@@ -0,0 +1,160 @@
...
+def extract_patches(tmp_specfile, patchqueue_filters, patchqueue_path,
+                    tmp_sources, target):
+    """
+    Extract patches and inject them into the specfile
+    """
+    for line in fileinput.input(tmp_specfile, inplace=True):
+        if any([ext in line for ext in patchqueue_filters]):
+            tar = tarfile.open(patchqueue_path)
+            for mem in tar.getmembers():
+                mem.name = '%s-' % target + os.path.basename(mem.name)
```
euanh: Could you please add a comment above this line to say that you are modifying mem.name in place to change the name of the extracted file (and drop any leading paths)?

planex/makesrpm.py
```diff
@@ -0,0 +1,157 @@
...
+            for mem in tar.getmembers():
+                # Modify mem.name in place to change the name of the
+                # extracted file (and drop any leading paths)
+                mem.name = '%s-' % target + os.path.basename(mem.name)
+                tar.extract(mem, tmp_sources)
+            patches = parse_patchseries(os.path.join(tmp_sources,
+                                                     '%s-series' % target))
+            print "# Patches for %s" % target
+            for patch_num, patch in enumerate(patches):
+                if patch != "":
```
euanh: This can just be `if patch:`

planex/makesrpm.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    Parse command line options
+    """
+    parser = argparse.ArgumentParser(description='Planex SRPM building')
+    add_common_parser_options(parser)
+    parser.add_argument(
+        "--topdir", metavar="DIR", default=None,
+        help='Set rpmbuild toplevel directory')
+    parser.add_argument(
+        "--dist", metavar="DIST", default="",
```
euanh: The default should be `None`, as for topdir.

planex/makesrpm.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    Return rpmbuild command line and arguments
+    """
+    cmd = ['rpmbuild']
+    if intercepted_args.quiet:
+        cmd.append('--quiet')
+    if intercepted_args.topdir is not None:
+        cmd.append('--define')
+        cmd.append('_topdir %s' % intercepted_args.topdir)
+    if intercepted_args.dist != "":
```
euanh: If you change the default above, this should become `if intercepted_args.dist is not None:`

salvocambria:
Fixed!

PR #262, opened 2016-04-14
debian: Remove experimental Debian package generation

euanh:
This code has not been maintained for a while and has fallen behind
the core RPM-building code.


euanh:
The debian code is still available in the debian branch of this repository.

PR #263, opened 2016-04-18
CA-202375: Build stunnel-4.15 against openssl-xs-devel-1.0.1e

phusl:
Signed-off-by: Phus Lu <phus.lu@citrix.com>

### DIFF
--- a/mk/Makefile
+++ b/mk/Makefile
@@ -18,6 +18,12 @@ buildrpms: conf_dir srpmutil
 	@cd .. && python setup.py install
 	echo confdir $(CONFIG_DIR)
 	chown -R root.root $(CONFIG_DIR)/*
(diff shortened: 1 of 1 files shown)


phusl:
Related with https://github.com/xenserver/xen-api-backports/pull/22

thomassa:
@phusl, do we still need this (and xenserver/xen-api-backports#22), or is it redundant if we are going to use the stunnel from the creedence-bugfix branch with the sanibel XenServer?

(If we do need it, I think this code-change looks fine.)

robhoes:
@phusl Could you please answer Thomas' question? Thanks!

phusl:
Sorry for late response @thomassa @robhoes 

Yes, I think we need this fix for sanibel-bugfix.

@thomassa understanding is right, if we use the stunnel from the creedence-bugfix, we do not need this fix.

But because this fix was already commited in sanibel-lcm, so I'd like to merge this fix to sanibel-bugfix branch also. :)

PR #268, opened 2016-05-06
mksrpm

salvocambria:
Obtain topdir name from tarball and write it inside the component specfile.


euanh:
@salvocambria Could you please rebase and update this pull request?   It won't apply at the moment.

euanh:
Merged manually in change 5e3ca2

Issue #271, opened 2016-05-10
planex-make-srpm does not handle missing arguments gracefully


euanh:
```
[root@ec799f010644 /]# planex-make-srpm 
Traceback (most recent call last):
  File "/usr/bin/planex-make-srpm", line 9, in <module>
    load_entry_point('planex==0.7.3-1-s0-0.7.3-16-g7150042.el7.centos', 'console_scripts', 'planex-make-srpm')()
  File "/usr/lib/python2.7/site-packages/planex/makesrpm.py", line 170, in _main
    main(sys.argv[1:])
  File "/usr/lib/python2.7/site-packages/planex/makesrpm.py", line 129, in main
    target = os.path.splitext(os.path.basename(passthrough_args[0]))[0]
IndexError: list index out of range
```

PR #272, opened 2016-05-10
mksrpm: handle missing arguments

salvocambria:
Exit with error message when no component specfile given as parameter to planex-make-srpm.


euanh:
@salvocambria: You need to rebase your mksrpm branch against master to get rid of all those extra changesets.   It looks like you have merged instead at some point.   The easiest thing to do might be to take a new branch from master (make sure you are up to date) and then cherry pick your bug fix onto it.

salvocambria:
It should be better now. I needed to rebase by fixing some stuff manually.

euanh:
Fixes #271

PR #278, opened 2016-06-13
Allow multiple pins from single repo

phusl:
Currently we cannot handle below pin file
```
{
  "SPECS/citrix-wlb.spec": {
    "0": "../wlb#master"
  },
  "SPECS/oem-wlb.spec": {
    "0": "../wlb#master"
  }
}
```
planex-pin will generate a single source tabball (i.e. _build/PINS/wlb-revision.tar.gz)
This behavior will cause `%setup -q -n wlb-%{version}` failure.

This patch adds origin_version as the prefix to generate different sources tarballs.

### DIFF
--- a/planex/pin.py
+++ b/planex/pin.py
@@ -183,8 +183,9 @@ def update(args):
 
             tmpdir = tempfile.mkdtemp(prefix='planex-pin')
             tmp_archive = archive(repo, treeish, orig_version, tmpdir)
-            tar_name = os.path.basename(tmp_archive).replace(orig_version,
-                                                             src_version)
+            tar_name = os.path.basename(tmp_archive)
+            tar_name = tar_name.replace(orig_version,
+                                        orig_version+'-'+src_version)
             tar_path = os.path.join(args.pins_dir, tar_name)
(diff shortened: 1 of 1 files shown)


euanh:
I'm not sure this is the best way to handle this problem.   Why can't you just have different %setup rules in your two spec files?  If both packages are built from exactly the same repo and tag, then it seems better to generate only one tarball.

Could you please post the output of the build failure you get when your patch is not applied?

Also, pylint and PEP-8 don't accept your whitespace changes - could you please fix those?

phusl:
Hi @euanh Thank you for your comments.

I support the opinion we shall use a single tarball if we archive it from same repo.

But the inside paths in archived tarball contains the `orig_version` prefix -- which we parsed it out from spec files.

In citrix-wlb.spec, the orig_version usual same as XenServer (i.e. 7.0.0)
But in oem-wlb.spec, the orig_version usual a specify number (i.e. 5.0.0)

OEM customer will surprise if they get a SPRM contains wlb-123.tar.gz/wlb-7.0.0/xxxx paths.

So the another solution maybe we could remove version info from `tar_name`

PR #280, opened 2016-06-21
Makefile.rules: echo important commands if VERBOSE is defined

srowe:


planex/Makefile.rules
```diff
@@ -34,6 +34,10 @@ DEPEND ?= planex-depend
 DEPEND_FLAGS ?= -d $(DIST) --topdir $(TOPDIR) --ignore-from ignore \
                 --pins-dir $(PINSDIR) $(DEPEND_EXTRA_FLAGS)
 
+ifndef VERBOSE
```
euanh: Could you please make this depend on the QUIET flag instead of adding a new flag?

PR #283, opened 2016-06-24
makesrpm: handle sources with no top-level directory

srowe:
Bitbucket generates source archives without a directory, adjust
%autosetup accordingly.


simonjbeaumont:
Looks OK to me... @euanh can we merge this one? Been sitting here for a few days.

PR #285, opened 2016-07-08
makesrpm: add initial support for patch guards

srowe:
Add a guard parameter to parse_patchseries() and use it to select
guarded patches. No caller defines it yet so it causes no positive
guards and all negative guards to be applied.

### DIFF
--- a/planex/makesrpm.py
+++ b/planex/makesrpm.py
@@ -4,6 +4,7 @@
 planex-make-srpm: Wrapper around rpmbuild
 """
 
+import re
 import sys
 import subprocess
 import os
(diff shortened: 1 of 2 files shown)


srowe:
Jira ticket is CA-215582

planex/makesrpm.py
```diff
@@ -33,16 +34,27 @@ def parse_args_or_exit(argv):
     return parser.parse_known_args(argv)
 
 
-def parse_patchseries(series_file):
+def parse_patchseries(series_file, guard = ''):
     """
     Parse series file and return the list of patches
     """
+    guard_re = re.compile(r'([\S]+)( #.*)?')
```
lindig: Would it be better to use `\s` instead of a space before `#`?

srowe: Possibly, I'm struggling to find a formal definition of the patch guard format. The convention is a single space but handling any whitespace would be prudent.

euanh:
@srowe: Could you please fix the pylint/pep8 errors?

srowe:
There are still functional issues with this. Let me try again.

planex/makesrpm.py
```diff
@@ -33,16 +34,35 @@ def parse_args_or_exit(argv):
...
+            line = line.strip()
+            if line and not line.startswith('#'):
+                match = guard_re.match(line)
+                if match.group(2):
+                    # guard present
+                    gtype = match.group(2)[2]
+                    guard_patch = match.group(2)[3:]
+
+                    if gtype == '+':
+                        if guard == guard_patch:
```
euanh: Could you please collapse this inner if into the outer one?

planex/makesrpm.py
```diff
@@ -33,16 +34,35 @@ def parse_args_or_exit(argv):
...
+                            # positive guard matches
+                            patches.append(match.group(1))
+                    elif gtype == '-':
+                        if guard != guard_patch:
```
euanh: Could you please collapse this inner if into the outer one?

euanh:
I see.   Could you please add a unit test for this function?   To make that easier it might be best to re-cast the function as a filter taking an iterable and returning another one.

I would still prefer to flatten the structure of the logic, something like this:
```
for line in series:
   if not line or line.startswith('#'):
      continue

   match = guard_re.match(line)
   if match.group(2):
      gtype = match.group(2)[2]
      guard_patch = match.group(2)[3:]

      if gtype == '+' and guard != guard.patch:
         continue
      if gtype == '-' and guard == guard.patch:
         continue

   yield match.group(1)
```

PR #286, opened 2016-07-12
depend: add Emacs and vim mode line

srowe:
The dependency recipes are more readable if editors are informed that
the content of the file is a Makefile.


srowe:
Oops, that's not in the right place...

euanh:
If you want this, you'll have to add the vim modeline, too :)

euanh:
@srowe: Could you please update the commit title?

PR #292, opened 2016-07-28
Generate dependencies from $(TOPDIR)/SPECS

srowe:
Make the dependency rules refer to $(TOPDIR)/SPECS rather than SPECS,
at present the directories are symlinked but this will cope with
$(TOPDIR)/SPECS containing links and/or real specfiles.


lindig:
Looks good to me.

PR #293, opened 2016-08-02
spec: Add function to return the highest numbered patch

srowe:


srowe:
I'll pull this into the other changes.

PR #301, opened 2016-08-09
Remove unused debianization code

euanh:
Remove leftover code for supporting automatic Debian package conversion.   All the original code is still available in the debian branch.
Removing this code allows us to tighten up the pylint configuration.


lindig:
The code in rpmextra doesn't look Debian specific - this was a little surprise for me as an outsider.

euanh:
@lindig: It was only used to support the script which made debs from rpms.

lindig:
> One of my most productive days was throwing away 1000 lines of code. — Ken Thompson

In this light: looks good to me.

lindig:
Time to :shipit: before more changes pile on.

PR #302, opened 2016-08-10
Add new top-level Dockerfile which rebuilds correctly when code changes

euanh:
This new Dockerfile builds a container based on the currently checked-out Git repository.  It has the following improvements over the old docker/Dockerfile.master:
 * It is in the top level of the repository, so a container can be built from a remote repository: docker build github.com/xenserver/planex
 * It works around a problem with running yum on OverlayFS-backed containers
 * It reduces the number of layers in the container
 * It is structured so that the container will be rebuilt correctly if the code in the local directory changes, but Docker's cache will be used as much as possible to cache earlier package installation steps


lindig:
I notice that you are using `&&` in the `Dockerfile` to reduce the number of layers. It's a little inconvenient to write but good for performance. So: thumbs up. This looks clean overall.

euanh:
Seems to be the recommended way to do it: 
 * https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/add-or-copy
 * https://www.ctl.io/developers/blog/post/more-docker-image-cache-tips/

lindig:
It is. It's just awkward that Docker's `RUN` semantics requires this when really Docker should better support tying commands together into one layer.

PR #303, opened 2016-08-12
Correctly return the name of the top level directory of the tarball

salvocambria:
makesrpm.py: when no topdir is present in the tarball, a simple 'os.path.commonprefix' would fail if all the files start with a common pattern.

extract.py: there can be multiple directories inside the tarball. It is wrong to simply assume that the first one of them is a topdir. This is correct only if there is a single file in the top level of the tarball.

### DIFF
--- a/planex/extract.py
+++ b/planex/extract.py
@@ -98,7 +98,7 @@ def archive_root(tar):
     """
     Return the name of the top level directory of the tarball
     """
-    if tar.firstmember.type == tarfile.DIRTYPE:
+    if len(tar.members) == 1 and tar.members[0].isdir():
         return tar.firstmember.name
     return ''
 
--- a/planex/extract.py
+++ b/planex/extract.py
@@ -98,6 +98,7 @@ def archive_root(tar):
     """
     Return the name of the top level directory of the tarball
     """
+    _ = tar.getnames()
     if len(tar.members) == 1 and tar.members[0].isdir():
         return tar.firstmember.name
     return ''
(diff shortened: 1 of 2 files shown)


planex/extract.py
```diff
@@ -98,7 +98,7 @@ def archive_root(tar):
     """
     Return the name of the top level directory of the tarball
     """
-    if tar.firstmember.type == tarfile.DIRTYPE:
+    if len(tar.members) == 1 and tar.members[0].isdir():
```
lindig: I'm surprised about `len(tar.members) == 1`. Does this mean the tar can only have one entry? Other tars still could have a top-level directory that this function would not return.

salvocambria: What we want to differentiate here is the following:
- tarball with one top level directory which contains all the sources
- tarball in which the sources are splattered at the top level without any directory.
We can only have these two cases.

An example of the first is:
tarball: test.tar
content: 
    test/a
    test/b
Instead, the second case looks like:
tarball: test.tar
content: a, b

salvocambria:
We need to run tar.getnames() before tar.members in order to get the full list of members. Without this, the tar.members will always return 1.

planex/extract.py
lindig: A third case is an empty tar. A tar can also contain symlinks. You might want to catch those.

salvocambria: Do we have components which have tarball of that kind? I am not sure that the entire logic would work with symlinks but if that is the case, and we can identify components which present such tarball, it is worth to test it.

lindig: I don't know but you might want to catch those cases with an assertion.

salvocambria:
Please ignore this pull request. Just realising that tar.members does not do what I was expecting.

PR #307, opened 2016-08-16
Correctly return topdir of tarball

salvocambria:
This fix will correctly return the topdir of a tarball (if any). Without this fix the code would return an incorrect result in situations in which all the files in the tarball have a common prefix but they are not stored in any topdir.

### DIFF
--- a/planex/extract.py
+++ b/planex/extract.py
@@ -98,8 +98,11 @@ def archive_root(tar):
     """
     Return the name of the top level directory of the tarball
     """
-    if tar.firstmember.type == tarfile.DIRTYPE:
-        return tar.firstmember.name
+    topname = os.path.commonprefix(tar.getnames())
+    if topname in tar.getnames():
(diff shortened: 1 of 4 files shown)


planex/extract.py
```diff
@@ -98,8 +98,11 @@ def archive_root(tar):
     """
     Return the name of the top level directory of the tarball
     """
-    if tar.firstmember.type == tarfile.DIRTYPE:
-        return tar.firstmember.name
+    topname = os.path.commonprefix(tar.getnames())
```
lindig: This all looks good. My only comment would be that you are calling `tar.getnames() twice. If this is an expensive operation it would make sense to store the result in a variable.

lindig:
I'd say :shipit:

salvocambria:
Some centos components (like systemd) have .tar.xz in their sources.

thomassa:
Looks fine to me as well. Merging.

PR #310, opened 2016-08-31
Per package sourcedirs

euanh:
Use per-package directories under SOURCES to store source files.  This removes the risk of name collisions between patches for different packages.

### DIFF
--- a/planex/Makefile.rules
+++ b/planex/Makefile.rules
@@ -45,13 +45,13 @@ endif
 # Main targets
 ############################################################################
 
-.PHONY: all rpms srpms srpm_repo
+.PHONY: all rpms
 
 all: $(TOPDIR) rpms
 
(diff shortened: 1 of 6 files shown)


planex/Makefile.rules
```diff
@@ -45,13 +45,13 @@ endif
 # Main targets
 ############################################################################
 
-.PHONY: all rpms srpms srpm_repo
+.phony: all rpms
```
lindig: According to my documentation of GNU Make, this should be `.PHONY:`

euanh: Thanks.  I didn't mean to change that!

lindig:
Besides the comment about `.PHONY` I don't have any other so far. The code looks good but I'm not a Python expert.

PR #316, opened 2016-09-21
spec: report fatal errors parsing spec file

srowe:
f70e6a discarded all output to stderr, including the cause of
exceptions raised by the rpm module. Redirect output to a file then
print any errors before re-raising exceptions.

### DIFF
--- a/planex/spec.py
+++ b/planex/spec.py
@@ -6,6 +6,9 @@
 import os
 import re
 import urlparse
+import sys
+import tempfile
+
 import rpm
 
 # Could have a decorator / context manager to set and unset all the RPM macros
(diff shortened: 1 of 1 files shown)


planex/spec.py
```diff
@@ -62,19 +64,25 @@ def __init__(self, path, dist="", check_package_name=True, topdir=None):
...
-        except ValueError as exn:
-            exn.args = (exn.args[0].rstrip() + ' ' + path, )
-            raise
+            except ValueError as exn:
+                nullfh.seek(0, os.SEEK_SET)
+                for line in nullfh:
+                    line = line.strip()
+                    if not line.endswith(': No such file or directory'):
+                        print >>sys.stderr, line
+                exn.args = (exn.args[0].rstrip() + ' ' + path, )
```
lindig: I'm not a Python expert, but `exn.args = (<expression>, )` looks strange with just nothing in the second component.

srowe: It's the way you define a tuple with a single value.

PR #317, opened 2016-09-22
repository: add class to process various source URL formats

srowe:


lindig:
For documentation, an example of the URLs being parsed could be added as a comment near the top of the respective parsing method.

srowe:
This changeset is very much work in progress, I think the supported URL formats should each have a unit test.

planex/repository.py
```diff
@@ -0,0 +1,82 @@
...
+            if query_path[1] == 'tags':
+                self.tag = query_path[2]
+            elif query_path[1] == 'heads':
+                self.branch = query_path[2]
+            else:
+                self.branch = 'master'
+        else:
+            self.branch = 'master'
+
+    def parse_hg(self):
```
euanh: Please rename this method to 'parse_gitweb'

planex/clone.py
```diff
@@ -0,0 +1,94 @@
+import argparse
+import json
+import logging
```
euanh: Unused import

planex/clone.py
```diff
@@ -0,0 +1,94 @@
...
+    args = parse_args_or_exit(argv)
+    setup_logging(args)
+
+    repos = checkout_remote_source(args.topdir, args.specfile)
+    if args.linkfile:
+        base_dir = os.path.join(args.topdir, repos[0].dir_name)
+        patch_dir = os.path.join(base_dir, '.git')
+        checkout_patchqueue(patch_dir, args.linkfile)
+        status = os.path.join(patch_dir, 'patches', repos[0].branch, 'status')
+        with open(status, 'w') as fileh:
```
euanh: Why are you opening this file?

planex/clone.py
```diff
@@ -0,0 +1,94 @@
+import argparse
```
euanh: Please add a short doc comment at the top of the file

planex/repository.py
```diff
@@ -0,0 +1,82 @@
+import logging
```
euanh: Please add a module doc comment

planex/repository.py
```diff
@@ -0,0 +1,82 @@
...
+import os.path
+import subprocess
+import urlparse
+
+
+class Repository(object):
```
euanh: Please add a class doc comment

planex/clone.py
srowe: Guilt needs this file to exist before it will allow operations on a patchqueue.

tests/test_repository.py
```diff
@@ -0,0 +1,55 @@
...
+
+import planex.repository
+
+
+class BasicTests(unittest.TestCase):
+    # unittest.TestCase has more methods than Pylint permits
+    # pylint: disable=R0904
+
+    def test_github(self):
+        url = "https://github.com/xapi-project/xen-api/archive/v1.10.1/xen-api-1.10.1.tar.gz"
```
srowe: Would you prefer they were moved to a JSON file and the tests read that?

planex/clone.py
srowe: The logical thing to use would be `guilt repair` but that barfs with "Branch master does not have a status file"

euanh: I thought it might be something like that.   It's fine, but please add a comment explaining it.

tests/test_repository.py
euanh: That would be much neater, if it's not too much trouble.

PR #333, opened 2016-10-12
Add a (very) preliminary planex-chroot command

dvrabel:
This command will start a planex container suitable for developer
builds of the listed packages.  The source for the packages must
already be available (e.g., with the planex-clone-sources command).


dvrabel:
Replacing this with a different method of building a docker container for each dev build.

PR #335, opened 2016-10-13
Docker: change "build" user creation to avoid scanning large directories

Pistahh:
Running the "usermod" command to change the uid of a user has the side
effect of changing the uid in all files in the home of the user to match
the new uid.

When ran in a directory having a large number of files on a slow fs (NFS)
it can take awfully long time.

The planex docker image uses usermod to set up the "build" user inside the
container with a right uid, so it has the issue described above.

With this change the "build" user is created with the right uid in the
first place, so "usermod" is not needed, thus fixing the performance issue.

Signed-off-by: Istvan Szekeres <istvan.szekeres@citrix.com>

### DIFF
--- a/docker/entry.sh
+++ b/docker/entry.sh
@@ -6,11 +6,18 @@ echo $@
 EXTUID=`stat -c %u /build`
 EXTGID=`stat -c %g /build`
 
-# Change 'build' UID in the container to match the owner of the
+# Create 'build' user in the container to match the owner of the
 # build directory, so that built packages will have the correct
 # owner outside the container.
-usermod --non-unique -u $EXTUID build
-groupmod --non-unique -g $EXTGID build
+groupadd build --gid $EXTGID      \
+               --non-unique
+useradd build --groups mock,wheel \
+              --home-dir /build   \
+              --uid $EXTUID       \
+              --gid $EXTGID       \
+              --no-create-home    \
+              --non-unique
+
 if [ -z "$1" ]; then
     exec su - build
 else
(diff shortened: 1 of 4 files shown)


docker/entry.sh
```diff
@@ -6,9 +6,16 @@ echo $@
...
-groupmod --non-unique -g $EXTGID build
+groupadd build --gid $EXTGID      \
+               --non-unique
+useradd build --groups mock,wheel \
+              --home-dir /build   \
+              --uid $EXTUID       \
+              --gid $EXTGID       \
+              --no-create-home    \
+              --non-unique
+
```
MarkSymsCtx: do we also need a chown build:build /build? Or do we get that as a consequence of mapping the external folder into the container?

Pistahh: Yes it is the consequence of the mapping and the usage of the right uid. And the purpose of the whole change is exactly to avoid the chown (done by usermod)

MarkSymsCtx: I did literally mean just the containing folder and not a recursive but so long as that gets the correct permissions as a result of the map then all is good.

euanh: The /build directory appears in the container with the uid and gid of the external user, so a chmod is not needed. The uid and gid of the build user come from running `stat` on /build

MarkSymsCtx: sweet, that's cool then

lindig: Is the basic idea to delay the creation of the `build` user until the container is started such that it then can be created with the right UID/GID? This is cool as long as the Dockerfile doesn't need the user earlier.

euanh:
@Pistahh This looks good.   There is now a conflict with another change in entry.sh - could you please rebase?

Pistahh:
rebased.

PR #336, opened 2016-10-14
planex-container: allow running an interactive shell

dvrabel:
Using the -c option runs the command without a controlling tty.  If no
arguments are specified, run an interactive shell that will have job
control etc.


lindig:
Looks good to me.

PR #338, opened 2016-11-03
repository: handle BitBucket URLs with hashes

srowe:
Fix the parsing of the query string when it does not start with
'refs/...'.

### DIFF
--- a/planex/repository.py
+++ b/planex/repository.py
@@ -138,18 +138,19 @@ class Repository(object):
         )
 
         self.dir_name = path[7]
-        query_str = urlparse.unquote(self.url.query)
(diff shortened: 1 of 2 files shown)


euanh:
Could you please update the tests to exercise this new code path?

planex/repository.py
```diff
@@ -143,13 +143,16 @@ def parse_bitbucket(self):
             query = query_str[3:]
             if '&' in query:
                 query = query[:query.find('&')]
-            query_path = query.split('/')
```
euanh: I think it would be better to use urlparse.parse_qs(query_str), which splits the query string into a dict.

PR #342, opened 2016-11-07
CA-228222 Handle commitish references with BitBucket

alexbrett:
This change adds support for converting git 'commitish' references in to a full SHA1 using the BitBucket API. This allows partial sha1's to be specified in a spec / lnk file, and the manifest generation to resolve them to full references.

Unit tests updated to cover this scenario, and tested in a local container.

Note adds a requirement for python-requests - spec file updated to cover this.

### DIFF
--- a/planex.spec
+++ b/planex.spec
@@ -14,6 +14,7 @@ Requires: mock
 Requires: python-argcomplete
 Requires: python-argparse
 Requires: python-setuptools
+Requires: python-requests
 Requires: rpm-build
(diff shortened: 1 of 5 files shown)


alexbrett:
Fixed up the various pylint warnings and modified commit messages. Note I have disabled the pylint 'too many attributes' check, as any other way of fixing this results (in my opinion anyway) in far less readable code, as it involves combing variables in a confusing way.

PR #344, opened 2016-11-09
Utility classes to help with handling tarballs, patchqueues and links

euanh:


planex/tarball.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def close(self):
+        """
+        Close the tarball
+        """
+        self.tarfile.close()
+
+    def extractfile(self, source):
+        """
+        Extract a file from the patchqueue, returning a file-like object
```
mseri: I think we should make this module more general and say `tarball` (instead of `patchqueue`) here and in the next docstring

PR #345, opened 2016-11-09
Rewrite linking to centralize all patch queue handling and rewriting in planex-make-srpm

euanh:
- The top level SOURCES directory is no longer symlinked in to _build, so source tarballs downloaded by planex-fetch are no longer mixed up with sources which come from the spec file repository.
 - patchqueues are no longer unpacked into SOURCES, removing the possibility of stale patches being left over from previous builds
 - All spec file manipulation now happens in planex-make-srpm's working directory.   The spec file in the SRPM is rewritten to include patches and to handle top level tarball directories which don't match rpmbuild's expectaitions.   The spec file in _build/SPECS is extracted from the patch queue unaltered and is only used by planex-depend.   planex-extract no longer rewrites files as it extracts them.

### DIFF
--- a/planex/depend.py
+++ b/planex/depend.py
@@ -31,14 +31,21 @@ def create_manifest_deps(spec):
     print '{}: {}'.format(manifest.get_path(spec_name), prereqs)
 
 
-def build_srpm_from_spec(spec):
+def build_srpm_from_spec(spec, lnk=False):
     """
     Generate rules to build SRPM from spec
     """
     srpmpath = spec.source_package_path()
-    print '%s: %s %s %s' % (srpmpath, spec.specpath(),
-                            " ".join(spec.source_paths()),
-                            manifest.get_path(spec.name()))
+    print '%s: %s' % (srpmpath, spec.specpath())
+    print '%s: %s' % (srpmpath, manifest.get_path(spec.name()))
+    for (url, path) in zip(spec.source_urls(), spec.source_paths()):
+        source = urlparse.urlparse(url)
+        if source.scheme in ["http", "https", "file", "ftp"]:
+            # Source was downloaded to _build/SOURCES
+            print '%s: %s' % (srpmpath, path)
+        elif not lnk:
+            # Source is local
+            print '%s: %s' % (srpmpath, "/".join(path.split("/")[1:]))
 
 
 def download_rpm_sources(spec):
@@ -128,6 +135,13 @@ def parse_cmdline():
     return parser.parse_args()
 
 
+def pkgname(path):
+    """
+    Return the name of the package at path
+    """
+    return os.path.splitext(os.path.basename(path))[0]
+
+
 def main():
     """
     Entry point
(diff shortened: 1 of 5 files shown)


planex/depend.py
```diff
@@ -167,7 +181,9 @@ def main():
                             defines=macros)
             pins[os.path.basename(pin_path)] = spec
 
-    for spec_path in args.specs:
+    links = {pkgname(lnk): lnk for lnk in args.specs if lnk.endswith(".lnk")}
+
+    for spec_path in [spec for spec in args.specs if spec.endswith(".spec")]:
```
mseri: I am fine enforcing lowercase extensions. Is this the case with all our specfiles?

planex/makesrpm.py
```diff
@@ -39,45 +43,24 @@ def parse_args_or_exit(argv):
...
-            topname = os.path.commonprefix(names)
-            if topname in names:
-                top_element = tar.getmember(topname)
-                if top_element.isdir():
-                    print "%s -n %s" % (line.strip(), topname)
-            else:
-                print "%s -c" % line.strip()
-        else:
-            print line,
+    return parsed_args
```
mseri: Given the amount of changes above, I think we should improve the helpstrings generated for the tool, spec argument and source argument. Or maybe add link and patches.tar as explicit arguments.

planex/makesrpm.py
```diff
@@ -39,45 +43,24 @@ def parse_args_or_exit(argv):
...
 
 
-def get_command_line(args, tmp_sources, tmp_specfile):
+def rpmbuild(args, tmpdir, specfile):
     """
-    Return rpmbuild command line and arguments
+    Run rpmbuild on working directory
```
mseri: We could specify that it returns the suprocess handle in the docstring.

planex/makesrpm.py
```diff
@@ -105,37 +88,54 @@ def main(argv):
...
 
     try:
-        # Copy files to temporary working area
-        copyfile(args.spec, tmp_specfile)
-        tarball_filters = ['.tar.gz', '.tar.bz2']
+        # Copy spec to working area
+        shutil.copyfile(args.spec, tmp_specfile)
 
+        # Copy sources to working area, rewriting spec as needed
+        tarball_filters = ['.tar.gz', '.tar.bz2']
```
mseri: We are implicitly assuming that all non-patchqueue tarballs are g/bzipped. Are we enforcing it somewhere?

euanh: This has always been the case in planex-make-srpm.   I'm not going to change it in this pull request.

planex/depend.py
euanh: Several other planex tools and the Makefile rules assume that the file extensions will be lowercase.   rpmbuild will create an SRPM containing foo.SPEC, but mock can't then build it.   Making the Makefile rules case insensitive is also not straightforward, so I think assuming lowercase is the best we can do.

euanh: I've tweaked the help text, but I think I'll be changing the interface a bit more in a future pull request.  I think it might be best to pick up the sources from their expected locations, as rpmbuild does, rather than passing them on the command line.  Also, it is redundant to pass a link file and a spec file at the same time - if a link file is passed, we can probably assume that we can fetch the spec file from its usual location or maybe just get it from the patchqueue tarball.

planex/makesrpm.py
mseri: I agree with the comment. I might spend some time try implementing it when I start working on the part of the dev-workflow that uses makesrpm.

PR #357, opened 2016-11-25
Replace planex-pin

euanh:
This pull request replaces planex-pin with a new, simpler implementation which can more easily be used to support pinning for linked packages.

Changes:
* Replaces the monolithic pins file with per-package files with the same format as link files
* Instead of rewriting source URLs in spec files, the new pin tool generates patch queue tarballs
* Pinned patch queue tarballs are handled by make-srpm in the same way as patch queue tarballs fetched via links

Missing:
* The new pinning tool does not currently generate meaningful package version numbers to indicate that patches have been applied


planex/cmd/pin.py
```diff
@@ -0,0 +1,74 @@
...
+import planex.git as git
+import planex.tarball as tarball
+import planex.util as util
+
+
+def parse_args_or_exit(argv=None):
+    """
+    Parse command line options
+    """
+    parser = argparse.ArgumentParser(description='Planex SRPM building')
```
mseri: Just a very minor thing. We should update the description as well.

jacusdb:
Please make the 3 changes that we agreed to in the workshop:

1. Rename planex-pin to planex-track
2. Use link files (instead of pin files with the same format)
3. Use an override directory

We need these changes in place before people start using the tool to avoid confusion.

euanh:
@jacusdb I've renamed planex-pin to planex-patchqueue.  I can't make the pin files have a .lnk extension in this pull request because Makefile.rules depends on this extension to cause the patchqueue to be built.   I'm hoping to address this in a better way in a future pull request.   The file formats are now identical and they are parsed by the same utility class.

PR #365, opened 2016-12-20
Override generation tool

mseri:
This PR introduces a first version of `planex-override`, a tool to automatically generate `.pin` files to override selected packages in `SPECS` with custom local repositories.

The tool is meant to be explicit, explaining the step performed and showing the generated output when invoked.

It might be worth to add ~~a `--dry-run` flag and~~ some unit tests as next step.

### DIFF
--- a/planex/cmd/override.py
+++ b/planex/cmd/override.py
@@ -27,9 +27,9 @@ def mkdir_p(path):
             raise exc
 
 
-def heuristical_is_spec_repo_root(xs_path):
+def heuristic_is_spec_repo_root(xs_path):
     """
-    Heuristical check for a spec repository root folder.
+    Heuristic check for a spec repository root folder.
     Looks for the presence of xs_path/.git, xs_path/SPECS and xs_path/mock.
     Raise and error if any of those is not present.
     """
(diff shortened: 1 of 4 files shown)


planex/cmd/override.py
```diff
@@ -0,0 +1,133 @@
...
+from planex.git import current_branch
+from planex.util import add_common_parser_options
+
+
+def mkdir_p(path):
+    """
+    Like `mkdir -p path` but does not fail when the path is already present.
+    """
+    try:
+        os.makedirs(path)
```
lindig: You explained another good reason: in a concurrent environment, there is a chance to create the directory between `isdir(path)` and `makedirs(path)`.

mseri: Missed that! Thanks

euanh: When we move to Python 3 we can just use `os.makedirs(..., exist_ok=True)` :)

lindig:
This has been in review for quite a while. Any reason not to merge it?

mseri:
Before merging, we are waiting for the new pinning work to be complete, and integrate any change if needed.

mseri:
I am going to close this PR and re-open it after the new changes on the pinning are landed on planex

PR #393, opened 2017-03-06
patchqueue: Fix patch generation with complex history

rosslagerwall:
git-format-patch is intended for use with simple linear histories which
means that it does not generate a consistent set of patches when the
history involves merges.  As an alternative, make a list of patches
similar to how git-format-patch does but include the diffs for merges
and ensure that the patch series is linear by following the first parent
only.

Signed-off-by: Ross Lagerwall <ross.lagerwall@citrix.com>

### DIFF
--- a/planex/git.py
+++ b/planex/git.py
@@ -97,6 +97,28 @@ def format_patch(repo, startref, endref, target_dir):
     dotgitdir = dotgitdir_of_path(repo)
 
     commit_range = "%s..%s" % (startref, endref)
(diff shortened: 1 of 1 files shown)


rosslagerwall:
If you think this is too ugly, see #394...

mseri:
I personally like #394 much more

rosslagerwall:
Yes, me too.

mseri:
I am closing this PR in favour of the other, but I want to wait @euanh opinion before doing anything with it

PR #394, opened 2017-03-06
patchqueue: Produce a single patch rather than a whole series

rosslagerwall:
Instead of using git-format-patch to produce a series of patches, use
git-diff to produce a single patch between the two refs. Using
git-format-patch is not appropriate because it is intended as a way of
emailing linear patch series and therefore doesn't work properly with
complex history.

Signed-off-by: Ross Lagerwall <ross.lagerwall@citrix.com>


euanh:
The format-patch method of building a patchqueue is going to be removed by the `pinning v3` project.

PR #395, opened 2017-03-07
Planex clone improvements

euanh:
* Make `planex-clone` handle structured branch names containing slashes.   This is needed for patchqueues, where the patchqueue must appear in a subdirectory matching the current branch name under .git/patches
* Make `planex-clone` create local branches when the upstream commitish does not point to a branch.

### DIFF
--- a/planex/cmd/clone.py
+++ b/planex/cmd/clone.py
@@ -31,6 +31,12 @@ def parse_args_or_exit(argv=None):
     return parser.parse_args(argv)
 
 
+def repo_name(url):
+    """Return the base repository name in url.   This is the name of
+       the directory which would be created by `git clone url`"""
+    return basename(url).rsplit(".git")[0]
+
+
 CHECKOUT_TEMPLATE = Template("""checkout poll: true,
          scm:[$$class: 'GitSCM',
               branches: [[name: '$branch']],
@@ -45,6 +51,7 @@ CHECKOUT_TEMPLATE = Template("""checkout poll: true,
 
 def clone_jenkins(url, destination, commitish, credentials):
     """Print Jenkinsfile fragment to clone repository"""
+    destination = join(destination, repo_name(url))
     print CHECKOUT_TEMPLATE.substitute(url=url,
                                        branch=commitish,
                                        checkoutdir=destination,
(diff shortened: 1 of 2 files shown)


planex/cmd/clone.py
```diff
@@ -76,10 +84,7 @@ def main(argv=None):
 
         if args.jenkins:
             print 'echo "Cloning %s"' % pin.url
-            print CHECKOUT_TEMPLATE.substitute(url=pin.url,
-                                               branch=pin.commitish,
-                                               checkoutdir=checkoutdir,
-                                               credentials=args.credentials)
+            clone_jenkins(pin.url, checkoutdir, pin.commitish, args.credentials)
```
mseri: I would be effect free in clone_jenkins, returning the fragment, and move the print back to this line.

planex/cmd/clone.py
```diff
@@ -79,34 +87,27 @@ def main(argv=None):
 
     for pinpath in args.pins:
         pin = Link(pinpath)
-        reponame = basename(pin.url).rsplit(".git")[0]
-        checkoutdir = join(args.repos, reponame)
 
         if args.jenkins:
             print 'echo "Cloning %s"' % pin.url
-            clone_jenkins(pin.url, checkoutdir, pin.commitish, args.credentials)
+            clone_jenkins(pin.url, args.repos, pin.commitish, args.credentials)
```
mseri: I am not suggesting to change it at this point, but I find annoying that `args.pins` and `args.repos` are simple 1 element variable but have a plural name.

mseri:
Just to understand, I am fairly neutral with respect to it, why did you decide to remove the support for running the script by hand?

planex/cmd/clone.py
euanh: They inherit their names from the command line arguments.  We could override the variable names, but this way it is clear which argument sets the variable.

planex/cmd/clone.py
euanh: clone() isn't side effect free, so I wanted to keep a similar signature for clone_jenkins()

euanh:
@mseri I removed the if __name__ == '__main__' to match all our other scripts.   You should only every be running these scripts through the setuptools wrapper.   Most of the scripts won't run at all unless the planex package has been installed.    For development the best way to work is to create a virtualenv and use `python setup.py develop`.

planex/cmd/clone.py
mseri: Yes, I know, and I am not suggesting to change them. It was just to point it out.

planex/cmd/clone.py
mseri: Makes sense, thanks.

PR #396, opened 2017-03-07
Pass --verbose if --quiet is not supplied so that logs are produced

MarkSymsCtx:
Log inside docker with concurrent builds and --output-sync=target


mseri:
Can you add `Signed-off-by...` to the commit message?

PR #397, opened 2017-03-07
Docker: Pass SSH_AUTH_SOCK through to sudo session in container

euanh:
Cloning repositories over ssh relies on having access to credentials.
Mount the SSH_AUTH_SOCK into the container as a volume and pass the
SSH_AUTH_SOCK environment variable through.   To do this, we need
to use sudo rather than su to become the build user, as sudo allows
us to set individual environment variables whereas for su environment
preservation is all-or-nothing.

### DIFF
--- a/docker/planex-container
+++ b/docker/planex-container
@@ -15,6 +15,8 @@ docker pull $PLANEX_CONTAINER
 docker run \
   --privileged \
   --rm -i -t \
+  -e SSH_AUTH_SOCK \
+  -v $(dirname $SSH_AUTH_SOCK):$(dirname $SSH_AUTH_SOCK) \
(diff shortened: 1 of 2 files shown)


docker/planex-container
```diff
@@ -15,6 +15,8 @@ docker pull $PLANEX_CONTAINER
 docker run \
   --privileged \
   --rm -i -t \
+  -e SSH_AUTH_SOCK \
+  -v $(dirname $SSH_AUTH_SOCK):$(dirname $SSH_AUTH_SOCK) \
```
srowe: Does this work when `SSH_AUTH_SOCK` is empty? `dirname ""` returns `.` which might not be what you want.

euanh:
Superseded by #401

PR #398, opened 2017-03-08
Add a check that a spec has %autosetup with a -p1 argument when applying patches

alexbrett:
When adding patch lines (e.g. Patch0 etc), planex assumes that they will be applied by rpm, but this is reliant on having a %autosetup with a -p1 argument in your %prep section.

This change verifies this assumption and raises an exception if it's not the case.

### DIFF
--- a/tests/test_patchqueue.py
+++ b/tests/test_patchqueue.py
@@ -41,9 +41,16 @@ class BasicTests(unittest.TestCase):
         self.assertIn("patch_with_a_negative_guard", applied)
 
     def test_rewrite_spec(self):
-        spec = Spec("tests/data/ocaml-uri.spec", check_package_name=False)
+        spec = Spec("tests/data/manifest/branding-xenserver.spec",
+                    check_package_name=False)
         patches = ["first.patch", "second.patch", "third.patch"]
         rewritten = planex.patchqueue.expand_patchqueue(spec, patches)
         self.assertIn("Patch0: first.patch\n", rewritten)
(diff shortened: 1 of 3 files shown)


lindig:
Just a remark: I had trouble with %autosetup mangling my spec file and had to move to %setup. The file can't be pinned and I'm fine if this is detected. However, it should still be possible to use the spec file unpinned. I can explain this also in person.

alexbrett:
That's fine - this check is only done when applying patches (ie when we are a patch queue or when pinning), specs just used as normal specs won't be affected.

tests/test_patchqueue.py
```diff
@@ -41,9 +41,16 @@ def test_patch_guards(self):
...
         rewritten = planex.patchqueue.expand_patchqueue(spec, patches)
         self.assertIn("Patch0: first.patch\n", rewritten)
         self.assertIn("Patch1: second.patch\n", rewritten)
         self.assertIn("Patch2: third.patch\n", rewritten)
+
+    def test_autosetup_check(self):
+        spec = Spec("tests/data/ocaml-uri.spec", check_package_name=False)
+        patches = ["first.patch"]
+        with self.assertRaises(planex.patchqueue.SpecMissingAutosetup):
+            planex.patchqueue.expand_patchqueue(spec, patches)
```
euanh: Could you add a negative test which shows that the exception is not raised for branding-xenserver?   That behaviour is implicit in test_rewrite_spec() but I think it would be better to have an explicit test.

PR #405, opened 2017-03-24
CP-21573

kostaslamda:


planex/cmd/mock.py
```diff
@@ -86,13 +89,20 @@ def createrepo(pkg_dir, metadata_dir, quiet=False):
...
+        extra_conf_list):
     """
     Write a new mock config, including a loopback repository configuration
     pointing to repo_path.    Ensure that the new config file's last-modified
     time is the same as the input file's, so that the mock chroot is not
     rebuilt.
     """
+    if extra_conf_list is None:
+        extra_conf_list = []
+
```
euanh: if you define the default in the argument parser to be [] then you don't need to make this check

kostaslamda: In Python, default values are created once when the function is defined, and are reused in each successive call. This means the if the object is mutable, like a `list`, mutations will persist through different invocations. In this case, we don't mutate the object, but I think it's consider bad practice.

euanh: I'm not suggesting adding a default to the function, I'm suggesting adding it to the argument parser:
```
argparser.add_argument("--loopback-config-extra", action="append", default=[], ...)
```

kostaslamda: Aaaaaaah, sure :)

PR #408, opened 2017-03-30
clone: Handle non-fatal errors more gracefully

euanh:
planex-clone currently exits if the destination directory for a repository already exists or the patchqueue is empty.   In the first case we should continue cloning other pins;  in the second case we should not fail when pushing an empty patchqueue.

### DIFF
--- a/planex/cmd/clone.py
+++ b/planex/cmd/clone.py
@@ -79,6 +79,31 @@ def clone(url, destination, commitish):
     return repo
 
 
+def apply_patchqueue(base_repo, pq_repo, pq_dir):
+    """
+    Apply a patchqueue to a base repository
+    """
+    # Symlink the patchqueue repository into .git/patches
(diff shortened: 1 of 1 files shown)


planex/cmd/clone.py
```diff
@@ -79,6 +79,31 @@ def clone(url, destination, commitish):
     return repo
 
 
+def apply_patchqueue(base_repo, pq_repo, pq_dir):
+    """
+    Apply a patchqueue to a base repository
+    """
+    # Symlink the patchqueue repository into .git/patches
+    link_path = relpath(pq_repo.working_dir, base_repo.git_dir)
```
mseri: As annoying as it sounds. I got errors for relative paths that are too nested while testing pins. I think here and below it would be safer to use absolute paths. I am trying to recall how I triggered the issue to give you a repro.

PR #409, opened 2017-04-03
Add manifest.json to SRPM if .gitarchive-info is present

MarkSymsCtx:


MarkSymsCtx:
Bug if the archive info isn't found, update coming.

PR #410, opened 2017-04-06
planex-create-mock-config: populate mock repos from yum config

srowe:
This tool creates a chroot configuration using a reference file and a
set of repositories that are defined in the yum configuration.

### DIFF
--- a/planex/cmd/createmockconfig.py
+++ b/planex/cmd/createmockconfig.py
@@ -1,0 +1,124 @@
+"""
+planex-create-mock-config: Populate mock config with yum repositories
+"""
+
+from __future__ import print_function
+import argparse
+import os
+import os.path
+import pprint
+import yum
+
+import ConfigParser
+import StringIO
+
+import argcomplete
+
+from planex.util import add_common_parser_options
+from planex.util import setup_logging
+from planex.util import setup_sigint_handler
+
+
+def load_mock_reference(fname):
+    """
+    read in the reference mock configuration
+
+    Chroot configurations are fragments of Python that mock `exec`s, mock goes
+    to some lengths to make this less insecure but they are rather convoluted.
+    We just trust that the caller has not included any surprises in the file.
+    """
+    config_opts = {}
+    execfile(fname)
+    return config_opts
+
+
+def load_yum_repos(repo_names):
+    """
+    read in the yum repository configuration
+    """
+    yum_base = yum.YumBase()
+    yum_repos = []
+    for repo_id in repo_names:
+        yum_repos += yum_base.repos.findRepos(repo_id)
+    return yum_repos
+
+
+def update_mock_repos(config, yum_repos):
+    """
+    Replace all repository sections with a new one derived from yum
+    """
+    for section in [s for s in config.sections() if s != 'main']:
+        config.remove_section(section)
+    for repo in yum_repos:
+        config.add_section(repo.id)
+        config.set(repo.id, 'name', repo.name)
+        if len(repo.baseurl):
+            config.set(repo.id, 'baseurl', ' '.join(repo.baseurl))
+        else:
+            config.set(repo.id, 'mirrorlist', repo.mirrorlist)
+        if repo.gpgcheck:
+            config.set(repo.id, 'gpgcheck', '1')
+            config.set(repo.id, 'gpgkey', ' '.join(repo.gpgkey))
+        else:
+            config.set(repo.id, 'gpgcheck', '0')
+
+
+def write_mock_cfg(fileh, cfg):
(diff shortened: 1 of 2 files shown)


planex/cmd/createmockconfig.py
```diff
@@ -0,0 +1,123 @@
...
+import argcomplete
+
+from planex.util import add_common_parser_options
+from planex.util import setup_logging
+from planex.util import setup_sigint_handler
+
+
+def load_mock_reference(fname):
+    """
+    read in the reference mock configuration
```
euanh: Could you please add a comment saying that this is the 'official' way to read the mock config, maybe with a reference to the mock code?

planex/cmd/createmockconfig.py
```diff
@@ -0,0 +1,123 @@
...
+
+def write_mock_cfg(fileh, cfg):
+    """
+    Save the mock configuration to a file handle
+    """
+    for key, value in cfg.items():
+        if key == 'yum.conf':
+            print('config_opts[\'%s\'] = """\n%s"""' % (key, value),
+                  file=fileh)
+        else:
```
euanh: This could be confused by a mock config with a `dnf.conf` section.   I think the `dnf.conf` format is the same as yum.conf: if so, then the 'yum.conf' branch should also handle `dnf.conf`; if not, you should probably bail out if you see `dnf.conf`.

srowe: The fedora rawhide config I have still uses `yum.conf`

    config_opts['package_manager'] = 'dnf'

    config_opts['yum.conf'] = """
    [main]

Do you have an actual Fedora host with mock I can check?

planex/cmd/createmockconfig.py
```diff
@@ -0,0 +1,123 @@
...
+    setup_sigint_handler()
+    args = parse_args_or_exit(argv)
+    setup_logging(args)
+
+    yum_repos = load_yum_repos(args.enablerepo)
+
+    # load the reference config
+    reference = os.path.join(args.configdir, args.root + '.cfg')
+    config_opts = load_mock_reference(reference)
+    mock_config_fp = StringIO.StringIO(config_opts['yum.conf'])
```
euanh: `yum.conf` might not be present if the target uses dnf.

planex/cmd/createmockconfig.py
euanh: `site-defaults.cfg` on my Fedora box has this to say:
```
# Contains content of $CHROOT/etc/yum/yum.conf or $CHROOT/etc/dnf/dnf.conf. If DNF is used and
# 'dnf.conf' is not set, then content of 'yum.conf' is used to populate $CHROOT/etc/dnf/dnf.conf
# config_opts['yum.conf'] = ''
# or
# config_opts['dnf.conf'] = ''
# Important! You must use 'assumeyes=1' in yum/dnf.conf otherwise Mock will fail.
```

planex/cmd/createmockconfig.py
srowe: The 'official' mock code is an abomination of `exec`, pipes and `pickle`. I wish I'd never seen it...

srowe: I guess using `execfile` is a security hole as you wedge

    subprocess.call(['rm', '-rf', '/'])

in the midst of your reference config file. Do we care?

planex/cmd/createmockconfig.py
```diff
@@ -0,0 +1,123 @@
...
+    update_mock_repos(mock_repos, yum_repos)
+    mock_config_fp.truncate(0)
+    mock_repos.write(mock_config_fp)
+    config_opts['yum.conf'] = mock_config_fp.getvalue()
+
+    # write new config
+    with open(args.mockconfig, "w") as fileh:
+        write_mock_cfg(fileh, config_opts)
+
+if __name__ == '__main__':
```
srowe: You seemed to have removed this from all the planex commands. Is there a reason, it makes it harder to dev test changes.

planex/cmd/createmockconfig.py
euanh: I don't think so, given that what mock itself does is similar.   I would just like a comment to reassure a future reviewer that it was done this way after careful consideration of the alternatives! :)

planex/cmd/createmockconfig.py
euanh: Yes, it's not needed if you use `python setup.py develop`, (you can do that in a `virtualenv` for better hygiene).   It may not be the case now, but in the past I remember running into problems when the scripts were run directly - perhaps to do with imports not being found.

PR #412, opened 2017-04-07
Rebuild pinned packages when the underlying repositories change

euanh:


planex/fileupdate.py
```diff
@@ -0,0 +1,62 @@
...
+
+        except IOError as ioe:
+            # Errno == 2 => output file does not exist
+            if ioe.errno != 2:
+                raise
+
+        print "Copying infile to outfile"
+        self.infile.seek(0)
+        with open(self.filename, 'wb') as outfile:
+            shutil.copyfileobj(self.infile, outfile)
```
mseri: why you decided to copy instead of move?

euanh: In this case I've used TemporaryFile so I don't have a filename and can't use shutil.move().   If I used TemporaryFile I could shutil.move() the file, but since /tmp is usually a separate filesystem the move would turn into a copy anyway.

planex/fileupdate.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+
+import hashlib
+import os
+import shutil
+import sys
+import tempfile
+
+def sameSize(infile, outfile):
+    infile.seek(0)
```
euanh: We need to sync(), otherwise os.fstat() sees a zero-length file:
```
$ python
Python 2.7.5 (default, Nov  6 2016, 00:28:07) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-11)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> infile = open("foo", "wb")
>>> infile.writelines("ntahusnthsnthstnh\n")
>>> import os
>>> os.fstat(infile.fileno()).st_size
0
```

Alternatively we can use tell(), but we will first need to seek() to the end of the file because we  the caller may have called seek() after writing to the file descriptor.

srowe:
shipit

PR #415, opened 2017-04-21
mock: ensure that build output is always visible

edwintorok:
mock behaves [differently](https://github.com/rpm-software-management/mock/blob/c59581ac78098b2b58b3a30a4a5d1f7338017e6b/py/mockbuild/util.py#L944) whether stderr is a tty or not.
`make output-sync=target` redirects the output, so that stderr is no longer a tty, causing the loss of build output.

Restore mock loglevel to normal, and use `pty.spawn` to turn mock's stderr into a pseudo-tty.
For compatibility with python2.7 explicitly check the exitcode of the child process, because we want to propagate build errors to the parent process.

Signed-off-by: Edwin Török <edvin.torok@citrix.com>

### DIFF
--- a/planex/cmd/mock.py
+++ b/planex/cmd/mock.py
@@ -3,6 +3,7 @@ planex-build-mock: Wrapper around mock
 """
 
 import os
+import pty
 import shutil
 import subprocess
 import sys
@@ -50,6 +51,31 @@ def parse_args_or_exit(argv=None):
     return parser.parse_args(argv)
 
 
+def pty_check_call(cmd):
+    """
+    Runs the given command in a subprocess with I/O redirected through a pty.
+    This ensures isatty(3) will return 1.
+    An exception is raised if the command exits with non-zero status.
+    """
+    # python2.7 doesn't return the exitcode here:
+    pty.spawn(cmd)
+    # get exit status of first child
+    (pid, status) = os.waitpid(-1, 0)
+    returncode = 1
+    if status == 0:
+        returncode = 0
+    elif os.WIFEXITED(status):
+        returncode = os.WEXITSTATUS(status)
+        print "PID %d exited with status %d" % (pid, returncode)
+    elif os.WIFSIGNALED(status):
+        signal = os.WTERMSIG(status)
+        print "PID %d exited with signal %d" % (pid, signal)
+    else:
+        print "PID %d exited with non-zero status 0x%02x" % (pid, status)
+    if returncode > 0:
+        raise subprocess.CalledProcessError(returncode, cmd)
+
+
 def mock(args, tmp_config_dir, *extra_params):
     """
     Return mock command line and arguments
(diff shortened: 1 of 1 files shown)


planex/cmd/mock.py
```diff
@@ -50,6 +51,22 @@ def parse_args_or_exit(argv=None):
...
+    pty.spawn(cmd)
+    # get exit status of first child
+    (pid, status) = os.waitpid(-1, 0)
+    if status != 0:
+        if os.WIFEXITED(status):
+            returncode = os.WEXITSTATUS(status)
+            print "PID %d exited with status %d" % (pid, returncode)
+        else:
+            # Killed by a signal, etc.
+            print "PID %d exited with non-zero status 0x%02x" % (pid, status)
```
srowe: For completeness can we have

    elif os.WIFSIGNALED(status)
        signal = os.WTERMSIG(status)
        print "PID %d exited with signal %d" % (pid, signal)

planex/cmd/mock.py
```diff
@@ -50,6 +51,22 @@ def parse_args_or_exit(argv=None):
     return parser.parse_args(argv)
 
 
+def pty_check_call(cmd):
+    # python2.7 doesn't return the exitcode here:
+    pty.spawn(cmd)
+    # get exit status of first child
+    (pid, status) = os.waitpid(-1, 0)
+    if status != 0:
+        if os.WIFEXITED(status):
```
srowe: I think these two `if`s should be the other way around.

planex/cmd/mock.py
edwintorok: sure, adding

planex/cmd/mock.py
edwintorok: I want to make sure I always raise an exception on non-zero exit status (both if program exits, gets killed, etc.). Would something like this look better?
```
+    returncode = 1
+    if status == 0:
+        returncode = 0
+    elif os.WIFEXITED(status):
+        returncode = os.WEXITSTATUS(status)
+        print "PID %d exited with status %d" % (pid, returncode)
+    elif os.WIFSIGNALED(status):
+        signal = os.WTERMSIG(status)
+        print "PID %d exited with signal %d" % (pid, signal)
+    else:
+        print "PID %d exited with non-zero status 0x%02x" % (pid, status)
+    if returncode > 0:
         raise subprocess.CalledProcessError(returncode, cmd)
```

srowe: Looks fine.

edwintorok: Thanks, force pushed the PR branch. (For the future/in general do you prefer force pushes for updating PRs under review, or extra commits with `git commit --fixup`/`git commit --squash`?)

MarkSymsCtx:
There is now a pylint error for a missing docstring on the new pty_check_call method, please add one.

edwintorok:
Thanks, added.

PR #421, opened 2017-05-10
Makefile.rules: Fetch rule potentially applies to everything in _build/SOURCES

euanh:
Previously we had an ugly list of suffixes matching files which could
be downloaded by planex-fetch.   It is neater to say that any target
in _build/SOURCES could potentially be downloaded.  This also allows
us to download non-archive sources which do not have a consistent set
of suffixes.


mseri:
This has been merged as part of https://github.com/xenserver/planex/pull/424

PR #422, opened 2017-05-10
Move the generation of .spec targets to the top-level Makefile

edwintorok:
This will allow the top-level Makefile to have more control
over the set of packages that are built.

`BUILD_TARGETS` will have to be defined in the top-level Makefile now (similarly to the template generated by the updated `planex-init`), and a missing variable is treated as a hard error.
(we could instead do the transition in two steps, first with just introducing `BUILD_TARGETS?=`, and another PR to actually move it)


euanh:
We ended up doing this differently in #425

PR #423, opened 2017-05-16
Remove preconfigured yum repos as last step of build 

MarkSymsCtx:
Avoid unexpected fallback by removing all yum repos and require the build process to add them

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,3 +34,5 @@ RUN python setup.py build \
 WORKDIR /build
 COPY docker/entry /entry
 ENTRYPOINT ["/entry"]
+
+RUN rm /etc/yum.repos.d/*


MarkSymsCtx:
We add our own repos from artifactory in the build process, we want those to be preferred and they should have everything needed.

euanh:
Yes, but we also use these containers for local builds, and being able to say `yum --enablerepo=base install <foo>` is very useful in that case.

MarkSymsCtx:
We have to avoid the names clashing then. Probably needs an update to the Jenkins build jobs to prefix or suffix the repo names before merging this

mseri:
I think an issue here is that the internal centos images are already using the correct snapshoted repositories, while the upstream centos is not

mseri:
@MarkSymsCtx @euanh Should we close or merge this PR?

PR #427, opened 2017-06-20
Patchqueue property tests

euanh:
This pull request adds some Hypothesis property-based tests for patchqueue guard handling.  Guard handling was already tested by point test in test_patchqueues.py so coverage reported by nosetests does not change, but the property-based tests explore some corner cases which the point test does not.

### DIFF
--- a/tests/test_patchqueue.py
+++ b/tests/test_patchqueue.py
@@ -1,22 +1,20 @@
 """Test patchqueue handling"""
 
-# Run these tests with 'nosetests':
-#   install the 'python-nose' package (Fedora/CentOS or Ubuntu)
-#   run 'nosetests' in the root of the repository
-
-import glob
-import os
-import os.path
 import shutil
-import sys
 import tempfile
 import unittest
+
+from hypothesis import given
+from nose.plugins.attrib import attr
+
+import tests.strategies as tst
 
 import planex.patchqueue
 from planex.spec import Spec
 
 
 class BasicTests(unittest.TestCase):
+    """Basic tests of patchqueue handling"""
     # unittest.TestCase has more methods than Pylint permits
     # pylint: disable=R0904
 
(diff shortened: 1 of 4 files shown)


tests/test_patchqueue.py
```diff
@@ -27,11 +27,11 @@ def tearDown(self):
         shutil.rmtree(self.test_dir)
 
     def test_patch_guards(self):
+        """Patch guards are applied correctly"""
         series = ("patch_no_guard_or_comment\n",
                   "patch_with_a_comment # this is a comment\n",
                   "patch_with_a_positive_guard #+aguard\n",
-                  "patch_with_a_negative_guard #-aguard\n",
-                  )
```
mseri: If you leave the parenthesis on a new line, adding a new patch will not require two line changes.

tests/test_patchqueue.py
```diff
@@ -50,13 +51,103 @@ def test_rewrite_spec(self):
         self.assertIn("Patch2: third.patch\n", rewritten)
 
     def test_autosetup_present(self):
+        """Patchqueue application succeeds if %autosetup is present"""
         spec = Spec("tests/data/manifest/branding-xenserver.spec",
                     check_package_name=False)
         patches = ["first.patch"]
-        planex.patchqueue.expand_patchqueue(spec, patches)
```
mseri: :O Was anything tested here before this PR?

mseri:
The strategies are impressively readable. We should start using hypothesis more.

mseri:
Feel free to merge or to add a newline before the parenthesis (see comment above)

tests/test_patchqueue.py
euanh: Yes, but this seems to be the style we use elsewhere.   (I noticed this because pylint complained that the closing parenthesis was wrongly indented.)

tests/test_patchqueue.py
euanh: Yes, this test was verifying that expand_patchqueue() did not raise an exception.   However pylint complained that the test did not use self and therefore didn't need to be a member.   This change seemed better than just saying 'self.assertTrue(True)' to shut pylint up...

PR #435, opened 2017-07-05
args: Define common argument parsers from which others can inherit

euanh:
argparse has a 'parents' mechanism which allows parsers to inherit options
from each other.   Use this instead of passing newly-created parsers to
a function which adds common options to them.


mseri:
Oh nice. We could add a default parser for `keeptmp` and one for `define` that are all over the place. And instead just pass them as parents.

PR #436, opened 2017-07-05
Makefile.rules: We only need one .DELETE_ON_ERROR target

euanh:
The existence of the .DELETE_ON_ERROR target is enough to cause make to
delete any target when the recipe used to produce it fails.   There is
no point in having several declarations of the target with different
prerequisites.

### DIFF
--- a/planex/Makefile.rules
+++ b/planex/Makefile.rules
@@ -62,6 +62,11 @@ endif
 # Main targets
 ############################################################################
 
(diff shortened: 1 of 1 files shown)


euanh:
@srowe Could you please sanity-check this?

srowe:
Yep, looks good. I need to apply this to my makefiles.

PR #442, opened 2017-08-02
[WIP] planex-pin: reintroduce to planex

mseri:
While cleaning up `planex-pin` to merge back into `planex`, I've realised that there are a few questions that need an answer to proceed.

You can find the code in a monolithic `cmd/pin.py` file in this PR and the issues underlined in `# TODO` comments.

Some of the helpers may be moved to the library, but I will keep them together until I have the final version.

### DIFF
--- a/planex/cmd/pin.py
+++ b/planex/cmd/pin.py
@@ -1,0 +1,233 @@
+"""
+planex-pin: generate pin files pointing at local repos
+in xenserver-specs/repos/ to override the spec/lnk
+"""
+from __future__ import print_function
+
+import argparse
+import json
+import os
+import sys
+from urlparse import urlunparse
+
+from planex.cmd.args import add_common_parser_options
+from planex.link import Link
+from planex.repository import Repository
+from planex.spec import Spec
+from planex.util import makedirs
+
+
+def heuristic_is_spec_repo_root(xs_path):
+    """
+    Heuristic check for a spec repository root folder.
+    Looks for the presence of xs_path/.git, xs_path/SPECS and xs_path/mock.
+    Raise and error if any of those is not present.
+    """
+
+    if not (os.path.exists("%s/.git" % xs_path) and
+                os.path.exists("%s/SPECS" % xs_path) and
+                os.path.exists("%s/mock" % xs_path)):
+        print("Spec repo not found in %s" % xs_path)
+        sys.exit(1)
+
+
+def spec_and_lnk(repo_path, package_name):
+    """
+    Return the Spec and Link object for
+    repo_path/SPECS/package_name.
+    Link can be None if not present.
+    Exception("package not present") otherwise
+    """
+    partial_file_path = "%s/SPECS/%s" % (repo_path, package_name)
+
+    specname = "%s.spec" % partial_file_path
+    if not os.path.isfile(specname):
+        print("Spec file for %s not present in %s/SPECS"
+              % (package_name, repo_path))
+        sys.exit(1)
+
+    spec = Spec(specname)
+
+    linkname = "%s.lnk" % partial_file_path
+    link = Link(linkname) if os.path.isfile(linkname) else None
+
+    return spec, link
+
+
+def repository_of(spec_or_link):
+    """Return the Repository of the provided Spec source url or Link url.
+       None if spec_or_link is None"""
+    if isinstance(spec_or_link, Spec):
+        return Repository(spec_or_link.source_urls()[0])
+    if isinstance(spec_or_link, Link):
+        return Repository(spec_or_link.url)
+    if spec_or_link is None:
+        return None
+    else:
+        print("repository_of: got unexpected object {}")
+        sys.exit(1)
+
+
+def repository_url(repo):
+    """
+    Return the first non-null value among repo.clone_url and
+    repo.url. If they are all None, it returns None.
+    """
+    if repo.clone_url is not None:
+        return repo.clone_url
+    if repo.url is not None:
+        return urlunparse(repo.url)
+
+
+def commitish_tag_or_branch(repo):
+    """
+    Return the first non-null value among the repository commitish,
+    tag or branch. If they are all None, it returns None.
+    """
+    if repo.commitish is not None:
+        return repo.commitish
+    if repo.tag is not None:
+        return repo.tag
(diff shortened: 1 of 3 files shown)


planex/cmd/pin.py
```diff
@@ -0,0 +1,233 @@
...
+from planex.repository import Repository
+from planex.spec import Spec
+from planex.util import makedirs
+
+
+def heuristic_is_spec_repo_root(xs_path):
+    """
+    Heuristic check for a spec repository root folder.
+    Looks for the presence of xs_path/.git, xs_path/SPECS and xs_path/mock.
+    Raise and error if any of those is not present.
```
euanh: Typo: raise and error

planex/cmd/pin.py
```diff
@@ -0,0 +1,233 @@
...
+    Parse command line options
+    """
+
+    parser = argparse.ArgumentParser(
+        description="Create a .pin file pointing to a repository "
+                    "in $CWD/repos. You must run "
+                    "this tool from the root of a spec repository.")
+    add_common_parser_options(parser)
+    parser.add_argument("package", help="package name")
+    parser.add_argument("--pinsdir", default="PINS",
```
euanh: I think we should print the new pin to stdout and let the user redirect it to the appropriate file.   If we do that, then we can get rid of the `--pinsdir` and `--dry-run` options and the code that goes with them.

planex/cmd/pin.py
```diff
@@ -0,0 +1,233 @@
...
+
+    return pinfile
+
+
+def make_pin(args, xs_path, package_name):
+    """
+    Return the pinfile path and a dict containing the content of the pinfile
+    """
+
+    # TODO: make the 'patchqueue' field customisable?
```
euanh: It would be nice to have.

planex/cmd/pin.py
```diff
@@ -0,0 +1,233 @@
...
+                    "in $CWD/repos. You must run "
+                    "this tool from the root of a spec repository.")
+    add_common_parser_options(parser)
+    parser.add_argument("package", help="package name")
+    parser.add_argument("--pinsdir", default="PINS",
+                        help="use custom pin folder (default to PINS)")
+    parser.add_argument("--url", metavar="URL", default=None,
+                        help="Source repository URL. It can be local "
+                             "e.g. repos/package)")
+    parser.add_argument("--branch", default=None,
```
euanh: Why not merge specifying the branch into `--url`, in the same way as you do for `--base` below?

planex/cmd/pin.py
```diff
@@ -0,0 +1,233 @@
...
+                             "the source tree to compile (defaults to "
+                             "master or the tag in the lnk file)")
+    parser.add_argument("--base", metavar="BASE#COMMITISH", default=None,
+                        help="Base repository URL and commitish. "
+                             "It can be local e.g. repos/package#master. "
+                             "This is used only for lnk packages.")
+    parser.add_argument("--dry-run", dest="dry", action="store_true",
+                        help="perform a dry-run only showing the "
+                             "performed steps")
+    parser.add_argument("--no-checks", dest="nochecks", action="store_true",
```
euanh: Let's get rid of this.

planex/cmd/pin.py
```diff
@@ -0,0 +1,233 @@
...
+        if args.base is not None:
+            base_url, base_commitish = args.base.split("#", 1)
+            base = base_url
+            base_commitish = base_commitish
+        else:
+            base = repository_url(base_repo)
+            # TODO: what whould we do if base is a url to a non-git tarball and the following is None?
+            base_commitish = commitish_tag_or_branch(base_repo)
+
+        # TODO: Should we take into account here if a spec is already in PINS?
```
euanh: Let's ignore this for now.

planex/cmd/pin.py
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    if link is not None:
+        if args.base is not None:
+            base_url, base_commitish = args.base.split("#", 1)
+            base = base_url
+            base_commitish = base_commitish
+        else:
+            base = repository_url(base_repo)
+            # TODO: what whould we do if base is a url to a non-git tarball and the following is None?
```
euanh: In this case I think you just don't need to write the base.

planex/cmd/pin.py
```diff
@@ -0,0 +1,233 @@
...
+from urlparse import urlunparse
+
+from planex.cmd.args import add_common_parser_options
+from planex.link import Link
+from planex.repository import Repository
+from planex.spec import Spec
+from planex.util import makedirs
+
+
+def heuristic_is_spec_repo_root(xs_path):
```
euanh: We can get rid of this along with the `--no-checks` option

planex/cmd/pin.py
```diff
@@ -0,0 +1,233 @@
...
+    if isinstance(spec_or_link, Link):
+        return Repository(spec_or_link.url)
+    if spec_or_link is None:
+        return None
+    else:
+        print("repository_of: got unexpected object {}")
+        sys.exit(1)
+
+
+def repository_url(repo):
```
euanh: Let's put this one into the repository class.

planex/cmd/pin.py
```diff
@@ -0,0 +1,233 @@
...
+    Return the first non-null value among repo.clone_url and
+    repo.url. If they are all None, it returns None.
+    """
+    if repo.clone_url is not None:
+        return repo.clone_url
+    if repo.url is not None:
+        return urlunparse(repo.url)
+
+
+def commitish_tag_or_branch(repo):
```
euanh: ...and this one

planex/cmd/pin.py
mseri: After finishing the changes, I remember why I had two fields. Most of the time you don't want to change the url but you want to customise the commitish. I could make it work with the new version by accepting a `--url #commitish`, as I am doing for `--base #commitish` for the repatched components, or we could separate them all.

mseri:
I have made the changes and rebased on master (I was on top of 0.23.1) to test it.
I am not very kin on using `#commitish` to modify only the commitish. I think we should be verbose and use `--url`, `--commitish`, `--base`, `--base_commitish`

PR #446, opened 2017-08-14
create-mock-config: Allow values in the [main] section of the

srowe:
'yum.conf' and 'dnf.conf' config options to be defined or overridden.

Signed-off-by: Simon Rowe <simon.rowe@citrix.com>

### DIFF
--- a/planex/cmd/createmockconfig.py
+++ b/planex/cmd/createmockconfig.py
@@ -63,7 +63,7 @@ def load_yum_repos(repo_names):
     return yum_repos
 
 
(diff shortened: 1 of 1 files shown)


MarkSymsCtx:
Is it worth adding an example to show how it could be used? e.g. reducing the verbosity/debug level of yum?

srowe:
Do you want the example in the commit message?

MarkSymsCtx:
I saw that you've done it on the internal ticket, that's good enough

srowe:
I've put some actual examples on the ticket.

PR #447, opened 2017-08-15
create-mock-config: convert config_opt values to native data types

srowe:


planex/cmd/createmockconfig.py
```diff
@@ -32,6 +32,15 @@ def __call__(self, parser, namespace, values, option_string=None):
         if '=' not in values:
             raise ValueError("invalid option: "+values)
         key, value = values.split('=', 1)
+
+        # convert to data types
+        if value == 'None':
+            value = None
+        if value in ('True', 'False'):
+            value = (value == 'True')
```
euanh: You can use bool(distutils.utils.strtobool(value)) here.   It will handle upper and lower case and alternatives such as 'yes' and 'no'.

https://stackoverflow.com/questions/715417/converting-from-a-string-to-boolean-in-python

srowe: Looks like `ast.literal_eval()` will handle all the data types

euanh: Yes, it should.

PR #449, opened 2017-08-17
planex-clone: Add option to skip cloning a patchqueue's base repository

euanh:
This new option allows the user to avoid cloning the base repository and applying the patchqueue to it.   This can be an expensive operation, and is not necessary if the user does not want to work on the package and instead just wants to build it.

### DIFF
--- a/planex/cmd/clone.py
+++ b/planex/cmd/clone.py
@@ -19,8 +19,6 @@ def parse_args_or_exit(argv=None):
     Parse command line options
     """
     parser = argparse.ArgumentParser(description='Clone sources')
-    parser.add_argument("-P", "--pins-dir", default="PINS",
-                        help="Directory containing pin overlays")
     parser.add_argument("--jenkins", action="store_true",
                         help="Print Jenkinsfile fragment")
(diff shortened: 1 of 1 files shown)


planex/cmd/clone.py
```diff
@@ -19,10 +19,10 @@ def parse_args_or_exit(argv=None):
     Parse command line options
     """
     parser = argparse.ArgumentParser(description='Clone sources')
-    parser.add_argument("-P", "--pins-dir", default="PINS",
-                        help="Directory containing pin overlays")
     parser.add_argument("--jenkins", action="store_true",
                         help="Print Jenkinsfile fragment")
+    parser.add_argument("--skip-base", default=False, action="store_true",
```
srowe: If you use `store_true` then the `default` is automatically `False` (and the other way round).

euanh: Yes, we've done this before.  I did it this way here so that the `if` condition below comes out as `if  not skip_base`, which is clearly related to the name of the command-line parameter for someone reading the code.   I'm not sure whether this is more valuable than avoiding negatives in the if statements, though...

PR #450, opened 2017-08-21
depend: Depend on local sources if link/pin does not provide them

euanh:
If a linked (or pinned )spec file refers to local sources and the link
file provides them, we do not need to print a dependency between the
SRPM and the source because this will be covered by the dependency on
the tarball built from the link.  If the link does not provide sources,
assume that they can be found in the specfile repository.

### DIFF
--- a/planex/cmd/depend.py
+++ b/planex/cmd/depend.py
@@ -33,7 +33,7 @@ def create_manifest_deps(spec):
                           manifest.get_path(spec.name())))
 
 
-def build_srpm_from_spec(spec, lnk=False):
+def build_srpm_from_spec(spec, lnk=None):
     """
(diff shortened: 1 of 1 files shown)


planex/cmd/depend.py
```diff
@@ -44,7 +44,10 @@ def build_srpm_from_spec(spec, lnk=False):
         if source.scheme in ["http", "https", "file", "ftp"]:
             # Source was downloaded to _build/SOURCES
             print('%s: %s' % (srpmpath, path))
-        elif not lnk:
+        elif lnk and lnk.sources:
+            # Use sources from patchqueue
+            pass
+        elif:
```
mseri: Shouldn't it be `else:`?

PR #454, opened 2017-08-23
Makefile.rules: Split up centralized _build/ directory creation

euanh:
This still creates the RPMS and MANIFESTS symlinks, but we plan to drop these in future.

### DIFF
--- a/planex/cmd/manifest.py
+++ b/planex/cmd/manifest.py
@@ -54,7 +54,7 @@ def parse_args_or_exit(argv=None):
 
 def get_path(package_name):
     """Get relative path to manifest file for package."""
-    return './MANIFESTS/{}.json'.format(package_name)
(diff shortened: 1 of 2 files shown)


planex/cmd/manifest.py
```diff
@@ -54,7 +54,7 @@ def parse_args_or_exit(argv=None):
 
 def get_path(package_name):
     """Get relative path to manifest file for package."""
-    return './MANIFESTS/{}.json'.format(package_name)
+    return '_build/MANIFESTS/{}.json'.format(package_name)
```
srowe: If `TOPDIR` is set to something else this will be wrong.

euanh: Noted, but it's not completely trivial to fix and we want to get rid of planex-manifest anyway.

PR #456, opened 2017-09-04
Allow to pass the location of the pin file by hand

mseri:

### DIFF
--- a/planex/cmd/pin.py
+++ b/planex/cmd/pin.py
@@ -147,6 +147,10 @@ def parse_args_or_exit(argv=None):
     parser.add_argument("--patchqueue", default="master",
                         help="Value for the patchqueue field of the pin file. "
                              "Defaults to master. ")
+    parser.add_argument("--to", default=None,
+                        help="Path of the pinfile to write. "
+                             "When used, it overwrites the file "
+                             "if present.")
     return parser.parse_args(argv)
 
 
@@ -158,7 +162,8 @@ def main(argv=None):
     args = parse_args_or_exit(argv)
 
     if args.base is not None and args.base_commitish is None:
-        print("Error: --base_commitish is required if --base is used.")
+        print("Error: --base_commitish is required if --base is used.",
+              file=sys.stderr)
         sys.exit(1)
 
     package_name = args.package
@@ -166,3 +171,12 @@ def main(argv=None):
     pin = make_pin(args, xs_path, package_name)
 
     print(json.dumps(pin, indent=2))
+
+    if args.to:
+        path = os.path.basename(args.to)
+        if os.path.exists(path):
+            with open(args.to, "w") as out:
+                json.dump(pin, out, indent=2)
+        else:
+            print("Error: path {} does not exist.".format(path),
+                  file=sys.stderr)


planex/cmd/pin.py
```diff
@@ -147,6 +147,10 @@ def parse_args_or_exit(argv=None):
     parser.add_argument("--patchqueue", default="master",
                         help="Value for the patchqueue field of the pin file. "
                              "Defaults to master. ")
+    parser.add_argument("--to", default=None,
```
euanh: Could you please call this `-o`/`--output`?   I think users are more likely to expect that than `--to`.

planex/cmd/pin.py
```diff
@@ -158,11 +162,21 @@ def main(argv=None):
...
 
     print(json.dumps(pin, indent=2))
+
+    if args.to:
+        path = os.path.basename(args.to)
+        if os.path.exists(path):
+            with open(args.to, "w") as out:
+                json.dump(pin, out, indent=2)
+        else:
+            print("Error: path {} does not exist.".format(path),
```
euanh: It might be better to use `sys.exit` here.   You can pass it an exit message instead of just an exit code.   If you still need to use print, please at least exit with a non-zero error code afterwards.

planex/cmd/pin.py
srowe: I think `None` is the default anyway.

planex/cmd/pin.py
```diff
@@ -158,11 +162,21 @@ def main(argv=None):
     args = parse_args_or_exit(argv)
 
     if args.base is not None and args.base_commitish is None:
-        print("Error: --base_commitish is required if --base is used.")
+        print("Error: --base_commitish is required if --base is used.",
```
srowe: Just use

    sys.exit("Error: ...")

planex/cmd/pin.py
```diff
@@ -158,11 +162,21 @@ def main(argv=None):
...
         sys.exit(1)
 
     package_name = args.package
     xs_path = os.getcwd()
     pin = make_pin(args, xs_path, package_name)
 
     print(json.dumps(pin, indent=2))
+
+    if args.to:
+        path = os.path.basename(args.to)
```
srowe: Did you mean to use `dirname()` here?

planex/cmd/pin.py
mseri: @euanh `-o` is good. I was wondering if it would be better to preserve the old behaviour instead of having to pass the output file.

@srowe true but I tend to be explicit to avoid breaking things if I move the flags to a an argument. I'd keep the explicit default if you don't mind

planex/cmd/pin.py
mseri: I did, thanks

mseri:
I have updated the PR

planex/cmd/pin.py
```diff
@@ -147,6 +146,10 @@ def parse_args_or_exit(argv=None):
     parser.add_argument("--patchqueue", default="master",
                         help="Value for the patchqueue field of the pin file. "
                              "Defaults to master. ")
+    parser.add_argument("--o", default=None,
```
srowe: `"-o", "--output"

mseri: Yes, I should have done it from the beginning, sorry. Fixed

PR #461, opened 2017-10-12
setup.py: remove include_package_data

srowe:
There is already an explicit map of files to include and defining both
prevents them from being included in a source distribution. See
https://stackoverflow.com/questions/7522250/how-to-include-package-data-with-setuptools-distribute

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,6 @@ from setuptools import setup
 setup(name='planex',
       version='2.1.0',
       packages=['planex', 'planex.cmd'],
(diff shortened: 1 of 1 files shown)


srowe:
The source dist contains the file now but `pip install` still doesn't install it...

euanh:
@srowe Is the outcome that `Makefile.rules` is in both the source and binary distributions now?

srowe:
I've only built and installed a source distribution.

PR #475, opened 2017-12-04
depend: Make a package depend on its Requires as well as its BuildRequires

### LINKED ISSUE #474: Ensure that a package's runtime dependencies are available before using it to build a downstream package
To build package **A**, `mock` needs to install all the package's build-time requirements (`BuildRequires`).   It does not need to install the package's runtime requirements (`Requires`).   However, to build package **B**, which depends on **A**, `mock` must install **A**'s _runtime_ requirements.   This causes problems if **A**'s runtime requirements come from the same spec repository as **A** itself.   `planex-depend` does not produce a Makefile dependency which will cause **A**'s runtime dependencies to be build before **B** is built, so it is the luck of the draw whether those packages exist at the time **B** is built.

Technically, `planex-depend` should make **B** have a build dependency on **A**'s runtime dependencies, however it is easier to make **A** depend on its own runtime dependencies, even though they are not strictly required in order to build it.

euanh:
To build package A, mock needs to install all the package's build-time
requirements (BuildRequires). It does not need to install the package's
runtime requirements (Requires). However, to build package B, which
depends on A, mock must install A's runtime requirements. This causes
problems if A's runtime requirements come from the same spec repository as
A itself. planex-depend does not produce a Makefile dependency which will
cause A's runtime dependencies to be build before B is built, so it is
the luck of the draw whether those packages exist at the time B is built.

Technically, planex-depend should make B have a build dependency on
A's runtime dependencies, however it is easier to make A depend on its
own runtime dependencies, even though they are not strictly required in
order to build it.

As with BuildRequires, if a runtime dependency is not provided by any
of the specs we are building, we will assume that the dependency comes
from the environment and no Makefile rule will be created.

Fixes #474

### DIFF
--- a/planex/cmd/depend.py
+++ b/planex/cmd/depend.py
@@ -102,7 +102,11 @@ def buildrequires_for_rpm(spec, provides_to_rpm):
     Generate build dependency rules between binary RPMs
     """
     rpmpath = spec.binary_package_paths()[-1]
-    for buildreq in spec.buildrequires():
+    # Package's Requires must exist for it to be installed as a
+    # BuildRequire of a later package, so we make it depend on
+    # Requires as well as BuildRequires to ensure they are built.
+    buildreqs = (spec.buildrequires() | spec.requires()) - spec.provides()
+    for buildreq in buildreqs:
         # Some buildrequires come from the system repository
         if buildreq in provides_to_rpm:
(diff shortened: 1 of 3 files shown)


mseri:
If nothing depends on A, can this unnecessarily build something?
Anyways, I think it's better to do a bit more than to miss some and fail the build. Overall this looks good

PR #476, opened 2017-12-07
CP-24858: Test building in travis-ci

makunterry:
1. Add 'disablerepo' option for planex-create-mock-config
2. Add building test for type N,RP,PQ repo


tests/specs/SPECS/N.spec
```diff
@@ -0,0 +1,129 @@
...
+%autosetup -p1 -n %{repo_name}-%{version}
+
+%build
+rm -rf %{buildroot}
+gopath=$(pwd)/gotools
+goroot=$gopath/go
+go_tool_name=go1.8.linux-amd64.tar.gz
+mkdir -p obj $gopath
+wget -O $gopath/$go_tool_name https://redirector.gvt1.com/edgedl/go/$go_tool_name
+tar -xzvf $gopath/$go_tool_name -C $gopath --overwrite
```
mseri: Can we avoid making this verbose aund use `tar xzf` instead? The output is huge

tests/specs/SPECS/N.spec
```diff
@@ -0,0 +1,129 @@
...
+%prep
+%autosetup -p1 -n %{repo_name}-%{version}
+
+%build
+rm -rf %{buildroot}
+gopath=$(pwd)/gotools
+goroot=$gopath/go
+go_tool_name=go1.8.linux-amd64.tar.gz
+mkdir -p obj $gopath
+wget -O $gopath/$go_tool_name https://redirector.gvt1.com/edgedl/go/$go_tool_name
```
mseri: Let's call `wget -q` this has unnecessary verbose and duplicated output.

tests/specs/SPECS/PQ.spec
```diff
@@ -0,0 +1,129 @@
...
+%prep
+%autosetup -p1 -n %{repo_name}-%{version}
+
+%build
+rm -rf %{buildroot}
+gopath=$(pwd)/gotools
+goroot=$gopath/go
+go_tool_name=go1.8.linux-amd64.tar.gz
+mkdir -p obj $gopath
+wget -O $gopath/$go_tool_name https://redirector.gvt1.com/edgedl/go/$go_tool_name
```
mseri: I think we should use `wget -q` (see above)

tests/specs/SPECS/PQ.spec
```diff
@@ -0,0 +1,129 @@
...
+tar -xzvf $gopath/$go_tool_name -C $gopath --overwrite
```
mseri: I think we should use `tar -xzf` (see above)

tests/specs/SPECS/RP.spec
```diff
@@ -0,0 +1,136 @@
...
+%patch2001 -p1
+%patch2002 -p1
+
+%build
+rm -rf %{buildroot}
+gopath=$(pwd)/gotools
+goroot=$gopath/go
+go_tool_name=go1.8.linux-amd64.tar.gz
+mkdir -p obj $gopath
+wget -O $gopath/$go_tool_name https://redirector.gvt1.com/edgedl/go/$go_tool_name
```
mseri: I think we should use `wget -q` (see above)

tests/specs/SPECS/RP.spec
```diff
@@ -0,0 +1,136 @@
...
+tar -xzvf $gopath/$go_tool_name -C $gopath --overwrite
```
mseri: I think we should use `tar -zxf` (see above)

planex/cmd/createmockconfig.py
```diff
@@ -53,15 +53,15 @@ def load_mock_reference(fname):
...
     """
     read in the yum repository configuration
     """
     yum_base = yum.YumBase()
     yum_repos = []
-    for repo_id in repo_names:
+    for repo_id in includes:
         yum_repos += yum_base.repos.findRepos(repo_id)
-    return yum_repos
+    return [rp for rp in yum_repos for ex in excludes if rp.id != ex]
```
mseri: I'd rewrite this more declaratively
```python
return [repo for repo in yum_repos if repo.id not in excludes]
```

I would ha preferred not touch this function at all and instead modify line 145-146 below as follows:

```
enablerepo = set(args.enablerepo) - set(args.disablerepo)
yum_repos = load_yum_repos(enablerepo)
```
but this cannot work with the wildcards as used in this code.

It still feels weird to me that we need to do that, couldn't we directly pass the right list of repos in the first place instead of passing the wildcard and then a disablerepo?

makunterry: Your first rewrite suggestion is very good, thanks!
And for the question, I prefer to follow the original one which can be passed as a wildcard so that we won't have to pass a list to planex-create-mock-config, cause it may cost too much effort to enumerate the whole list items.

makunterry: So for this point, is it OK with you about such kind of configuration to parameters?
wildcard enabled for `--enablerepo`, but wildcard disabled for `--disablerepo`.

euanh: Couldn't we do something like this?
```
set([yum_base.repos.findRepos(repo_id) for repo_id in args.enablerepo]) - set([yum_base.repos.findRepos(repo_id) for repo_id in args.disablerepo])
```

makunterry: Thanks @euanh for your comments.
And because each time calling `yum_base.repos.findRepos` it will return a list instead of item, so I made some minor changes in this code.

planex/cmd/createmockconfig.py
```diff
@@ -53,15 +53,18 @@ def load_mock_reference(fname):
...
+def load_yum_repos(includes, excludes):
     """
     read in the yum repository configuration
     """
     yum_base = yum.YumBase()
-    yum_repos = []
-    for repo_id in repo_names:
-        yum_repos += yum_base.repos.findRepos(repo_id)
-    return yum_repos
+    enablelist = []
```
euanh: Please use list (or set) comprehensions instead of loops here

makunterry: @euanh Refined. Thanks!

planex/cmd/createmockconfig.py
```diff
@@ -58,13 +58,11 @@ def load_yum_repos(includes, excludes):
...
     """
     yum_base = yum.YumBase()
-    enablelist = []
-    disablelist = []
-    for repo_id in includes:
-        enablelist += yum_base.repos.findRepos(repo_id)
-    for repo_id in excludes:
-        disablelist += yum_base.repos.findRepos(repo_id)
-    return list(set(enablelist) - set(disablelist))
+    enabled_set = {repo for pattern in includes
```
mseri: Nice!!

PR #482, opened 2017-12-14
CP-26319: Add None type check for enablerepo & disablerepo 

makunterry:
Signed-off-by: Kun Ma <kun.ma@citrix.com>

### DIFF
--- a/planex/cmd/createmockconfig.py
+++ b/planex/cmd/createmockconfig.py
@@ -119,9 +119,9 @@ def parse_args_or_exit(argv=None):
                         help="mock config directory")
     parser.add_argument("-r", "--root", metavar="INCFG", required=True,
                         help="reference chroot config")
-    parser.add_argument("--enablerepo", action="append",
+    parser.add_argument("--enablerepo", action="append", default=['*'],
                         help="Repository to include")
-    parser.add_argument("--disablerepo", action="append",
+    parser.add_argument("--disablerepo", action="append", default=[],
                         help="Repository to exclude")
     parser.add_argument("--config_opt", action=DictAction, metavar="OPT=VALUE",
                         help="Define mock configuration settings")
--- a/planex/cmd/createmockconfig.py
+++ b/planex/cmd/createmockconfig.py
@@ -98,8 +98,7 @@ def load_yum_repos(repo_config_list):
     read in the yum repository configuration
     """
     yum_base = yum.YumBase()
-    repo_set = {repo for repo in yum_base.repos.findRepos('*')
-                if repo.enabled}
+    repo_set = set(yum_base.repos.listEnabled())
     for config, pattern in repo_config_list:
         if config == "enable":
             repo_set |= set(yum_base.repos.findRepos(pattern))
--- a/planex/cmd/createmockconfig.py
+++ b/planex/cmd/createmockconfig.py
@@ -40,6 +40,46 @@ class DictAction(argparse.Action):
         setattr(namespace, self.dest, dictvalue)
 
 
+class RepoAction(argparse.Action):
+    """
+    Action subclass to form common function to add new repo config
+    """
+    # pylint: disable=R0903,W0223
+    def __init__(self, option_strings, dest, nargs=None, **kwargs):
+        if nargs is not None:
+            raise ValueError("nargs not allowed")
+        dest = "repo_config_list"
+        super(RepoAction, self).__init__(option_strings, dest, **kwargs)
+
+    def sub_call(self, namespace, values, repo_config='enable'):
+        """
+        A common function to implement __call__
+        """
+        repo_config_list = getattr(namespace, self.dest)
+        if repo_config_list is None:
+            repo_config_list = []
+        repo_config_list.append((repo_config, values))
+        setattr(namespace, self.dest, repo_config_list)
+
+
+class EnableRepoAction(RepoAction):
+    """
+    RepoAction subclass to form adding a enablement repo config
+    """
+    # pylint: disable=R0903
+    def __call__(self, parser, namespace, values, option_string=None):
+        self.sub_call(namespace, values, "enable")
+
+
+class DisableRepoAction(RepoAction):
+    """
+    RepoAction subclass to form adding a disablement repo config
+    """
+    # pylint: disable=R0903
+    def __call__(self, parser, namespace, values, option_string=None):
+        self.sub_call(namespace, values, "disable")
+
+
 def load_mock_reference(fname):
     """
     read in the reference mock configuration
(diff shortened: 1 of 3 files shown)


MarkSymsCtx:
Could you not just set defaults for the two arguments in the parse_args_or_exit method?

planex/cmd/createmockconfig.py
```diff
@@ -119,9 +119,9 @@ def parse_args_or_exit(argv=None):
                         help="mock config directory")
     parser.add_argument("-r", "--root", metavar="INCFG", required=True,
                         help="reference chroot config")
-    parser.add_argument("--enablerepo", action="append",
+    parser.add_argument("--enablerepo", action="append", default=['*'],
```
euanh: This will mean that '*' is always enabled.   The default for `enablerepo` should be an empty list, the same as for `disablerepo` below.  Then in `main` you can call `load_yum_repos` as follows:
```
load_yum_repos(args.enablerepo or ['*'], args.disablerepo)
```

MarkSymsCtx: I don't see how that is different. If enablerepo is supplied on the commandline it gets a value, if its not then it gets ['*'] both ways are equivalent in python AFAICT.

euanh: `action="append"` means that the command line argument is appended to the default, so the enabled list will always begin with `'*'`

euanh: However I'm not sure that enabling `'*'` when no argument is given is the right thing to do.   That will enable repositories which are disabled in the yum configuration.

euanh: I think we should start with the set of repos which are enabled in the yum configuration, then remove disabled repos and add enabled ones:
```
base_set = set(yum.repos.listEnabled())
enabled_set = {repo for pattern in includes
               for repo in yum_base.repos.findRepos(pattern)}
disabled_set = {repo for pattern in excludes
                for repo in yum_base.repos.findRepos(pattern)}
return list ((base_set - disabled_set) | enabled_set)
```

This has the following properties (if ``--enablerepo` and `--disablerepo` both default to []):
 * If `--enablerepo` and `--disablerepo` are omitted, we get the set of repos which are enabled in the yum configuration
 * If `--enablerepo='*'` is given, we get all repos in the yum configuration
 * If `--enabledrepo='foo'` is given, we get all repos in the yum configuration plus foo
 * If `--enablerepo='foo'` and `--disablerepo='*'` are given, we get only foo

This matches my understanding of the semantics of `yum` and `dnf`'s `--enablerepos` and `--disablerepos` arguments.

@srowe You wrote this tool originally - what semantics did you expect?

srowe: I use constructs like `--disablerepo='*' --enablerepo=foo` to enable only `foo`, I think we need to look at how `yum`/`dnf` process their options.

euanh: I think that's how yum works, and my suggestion above should do the same (the last bullet - tested in the Python toplevel).    I guess we need to check whether yum accumulates all `--enablerepo` and `--disablerepo` arguments into two sets or does something which respects the order in which they appear.   If it's the latter we would need to change our argument parser to tag `--enablerepo` and `--disablerepo` arguments and accumulate them in a list which we then walk.   (I don't see any benefit in it behaving that way - it's a lot of complexity for limited benefit.)

srowe: Order is significant

    [test@localhost ~]$ yum --disablerepo='*' --enablerepo=updates  repolist
    Loaded plugins: fastestmirror
    Repodata is over 2 weeks old. Install yum-cron? Or run: yum makecache fast
    Loading mirror speeds from cached hostfile
    repo id                                                     repo name                                                   status
    !updates                                                    updates                                                     1,241
    repolist: 1,241
    [test@localhost ~]$ yum --enablerepo=updates --disablerepo='*'  repolist
    Loaded plugins: fastestmirror
    Loading mirror speeds from cached hostfile
    repolist: 0

euanh: Yes, I looked at the code for yum-cli and it processes the arguments in order and applies them directly to the yumbase object. (it doesn't use argparse)

makunterry:
@srowe @euanh and @MarkSymsCtx 
I have refined the createmockconfig.py and added 3 Action class(1 base and 2 sub-classes) to handle the repo config.
Now with the code change it can handle the different order of parameters from `--disablerepo` and `--enablerepo`.
I tested on my docker container:
```
[root@aeb0715f30b8 test]# ./a.py --enablerepo 'updates' --disablerepo '*'
Namespace(repo_config_list=[('enable', 'updates'), ('disable', '*')])
-----------------
enable ==> updates
disable ==> *
-----------------
Enabled repos:
Loaded plugins: fastestmirror, ovl
[]
[root@aeb0715f30b8 test]# ./a.py --disablerepo '*' --enablerepo 'updates'
Namespace(repo_config_list=[('disable', '*'), ('enable', 'updates')])
-----------------
disable ==> *
enable ==> updates
-----------------
Enabled repos:
Loaded plugins: fastestmirror, ovl
['updates']
```
As we can see, the first time `--enablerepo` first and then `--disablerepo` all, the result is empty list.
The second time `--disablerepo` all first and then `--enablerepo "updates"`, the result is ["updates"].

And the pylint disables:
`W0223: Method %r is abstract in class %r but is not overridden`

planex/cmd/createmockconfig.py
```diff
@@ -53,16 +93,20 @@ def load_mock_reference(fname):
...
     """
     read in the yum repository configuration
     """
     yum_base = yum.YumBase()
-    enabled_set = {repo for pattern in includes
-                   for repo in yum_base.repos.findRepos(pattern)}
-    disabled_set = {repo for pattern in excludes
-                    for repo in yum_base.repos.findRepos(pattern)}
-    return list(enabled_set - disabled_set)
+    repo_set = {repo for repo in yum_base.repos.findRepos('*')
```
srowe: repo_set = set(repo ...

makunterry: @srowe This is set comprehension and will return a `set`, so we don't need to use a `set(...)` .

srowe: Thanks, I hadn't come across it.

euanh: I think you can replace this with:
```
set(yum_base.repos.listEnabled())
```

planex/cmd/createmockconfig.py
```diff
@@ -53,16 +93,20 @@ def load_mock_reference(fname):
...
+                if repo.enabled}
+    for config, pattern in repo_config_list:
+        if config == "enable":
+            repo_set |= set(yum_base.repos.findRepos(pattern))
+        elif config == "disable":
+            repo_set -= set(yum_base.repos.findRepos(pattern))
+
+    return list(repo_set)
```
euanh: This looks good.  Could you please write a couple of test cases to verify the behaviour?

makunterry: @euanh do you mean the unit test to cover those code changes?

euanh: Yes.  It might be a bit tricky.   It should be possible to use mock (the mocking library, not the RPM builder!) or some other technique to override the yum_base object and return dummy repos.   Then run your method with different sets of options and verify that it works as you expect.

MarkSymsCtx: That should work python mock can patch replace anything that you import.

makunterry: @euanh and @MarkSymsCtx 
Thanks.
I have updated the PR and create a test case for `createmockconfig.load_yum_repo`.
Please help to check.

PR #490, opened 2018-01-30
Consolidate Spec's various source getter methods

euanh:
This pull request consolidates most sources getters into sources() and adds a source() convenience method to fetch the source corresponding to a downloaded file name.

local_sources() and local_patches() remain but will be dealt with in the next pull request.


euanh:
My next pull request will replace the tuple with a class! :)

PR #495, opened 2018-02-15
pin: use sys.exit to also print the error message

mseri:


euanh:
Argh - lines too long. :(

Issue #501, opened 2018-03-27
All utilities must accept a spec file and optional link file as arguments


euanh:
For this to happen:
Depend must make the spec and link (if it exists) prerequisites of each target
Recipes in Makefile.rules must handle a spec and optional link (probably by using $^ instead of $<)

PR #513, opened 2018-03-27
fetch: Take spec file and an optional link as arguments

euanh:
All commands now take the spec file and an optional link as arguments.
This is necessary because the link can contain RPM macros which depend
on the spec file.

Fixes: #501


planex/Makefile.rules
```diff
@@ -101,7 +101,7 @@ $(TOPDIR)/MANIFESTS/%.json:
 $(TOPDIR)/SOURCES/%:
 	@echo [FETCH] $@
 	$(AT) mkdir -p $(@D)
-	$(AT)$(FETCH) $(FETCH_FLAGS) $< $@
+	$(AT)$(FETCH) $(FETCH_FLAGS) $^ $@
```
mseri: What does `$^` mean?

srowe: All dependencies, rather than the leftmost dependency

euanh: https://www.gnu.org/software/make/manual/make.html#Automatic-Variables

mseri: Thanks!

PR #514, opened 2018-03-28
Follow spec file sourcenames semantics in link files

### LINKED ISSUE #504: Follow specfile source name semantics in link files
The new link parsing logic in spec.py uses the resource names given in the link file verbatim.   If the link is overriding a source already defined in the spec file, the name in the link must match the name in the spec exactly otherwise the link -defined source will be added as an extra resource.

The spec file format allows the same source to be named in several different ways.  We should implement the same logic in link:

   * source, source0, source0000 are all the same
  * source 1, source001 are the same
 * patch, patch0, patch000 are the same
* ...

We should parse the integer at the end of the source name and use that as the index in the resoources dict in the spec class.   If the integer is missing, assume it is source 0.

mseri:
This should fix #504

### DIFF
--- a/planex/spec.py
+++ b/planex/spec.py
@@ -328,11 +328,11 @@ class Spec(object):
                         "spec file name '%s' does not match package name '%s'"
                         % (path, self.name()))
 
-        for source in reversed(self.spec.sources):
-            if source[2] == 1:
-                self.add_source("source%d" % source[1], source[0], path)
-            elif source[2] == 2:
-                self.add_patch("patch%d" % source[1], source[0], path)
+        for filepath, index, sourcetype in reversed(self.spec.sources):
+            if sourcetype == 1:
+                self.add_source(index, filepath, path)
+            elif sourcetype == 2:
+                self.add_patch(index, filepath, path)
 
     def specpath(self):
         """Return the path to the spec file"""
--- a/planex/spec.py
+++ b/planex/spec.py
@@ -86,6 +86,7 @@ def parse_spec_quietly(path):
 
 def expandmacros(func):
     """Decorator to expand RPM macros in strings"""
+
     def func_wrapper(self):
         """Decorator wrapper"""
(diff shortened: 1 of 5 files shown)


planex/spec.py
```diff
@@ -256,6 +258,29 @@ def series(self):
             return queue.series()
 
 
+def _parse_name(name):
+    """
+    Parse [name] a string composed by a workd and an optional number into the
```
euanh: workd -> word

planex/spec.py
```diff
@@ -256,6 +258,29 @@ def series(self):
...
+    Raises KeyError if [name] cannot be parsed
+    """
+    matcher = re.compile(r"\A[a-zA-Z]+(\d*)\Z")
+    match = matcher.match(name)
+
+    if match is None:
+        raise KeyError(name)
+
+    (idx, ) = match.groups()
+    if idx == 0:
```
euanh: Shouldn't this be `if idx == ''`?

planex/spec.py
srowe: "word"?

tests/test_spec.py
```diff
@@ -98,84 +98,84 @@ def test_resources(self):
...
                      "SOURCES/"),
-             Patchqueue(self.spec, "patchqueue0", "http://foo/patchqueue.tar",
+             Patchqueue(self.spec, "http://foo/patchqueue.tar",
                         "link1", "PATCHES/")])
 
     def test_resource(self):
         """URLs for individual resources are correct"""
         self.assertEqual(
             self.spec.resource("path/to/ocaml-cohttp-0.9.8.tar.gz"),
-            Blob(self.spec, "source0",
```
euanh: Could you please make one of these tests add 'source' with no digit, to test the special case in `_parse_name`

planex/spec.py
mseri: Yes, broken when rebasing, well spotted!

PR #516, opened 2018-03-28
link: Anchor parsing regexes to start and end of keys

euanh:
The change of the 'patch' name parsing regex to accept zero or more digits
(patch\d*) causes it also to match 'patchqueue'.

### DIFF
--- a/planex/link.py
+++ b/planex/link.py
@@ -56,7 +56,7 @@ class Link(object):
         if self.schema_version < 2:
             return self.link.get('sources', None)
 
-        patch_matcher = re.compile(r'source\d*', re.IGNORECASE)
+        patch_matcher = re.compile(r'^source\d*$', re.IGNORECASE)
         return {k: v for k, v
                 in self.link.iteritems()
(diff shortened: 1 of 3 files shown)


planex/link.py
```diff
@@ -56,7 +56,7 @@ def sources(self):
         if self.schema_version < 2:
             return self.link.get('sources', None)
 
-        patch_matcher = re.compile(r'source\d*', re.IGNORECASE)
+        patch_matcher = re.compile(r'\Asource\d*\Z', re.IGNORECASE)
```
srowe: `match()` already implies start of line.

srowe: Is `\Z` any clearer than `$`? (I had to look it up)

mseri: No, I was not sure if python re supported `$` and I remembered that it did support `\A` and `\Z`. I'm happy to change it.

mseri: Ops, I thought it was in my recently-merged PR. This is uniform with what is done in `spec.py`, if we change it here we should probably update it also there.

Issue #511, opened 2018-03-27
Support additional sources in links


euanh:
At present a source or patch defined in a link file can only override a source or patch of the same name which is already defined in the spec file.   It would be very useful to be able to add extra sources by listing them in the link file (although this would depend on the source unpacking logic in the %prep stanza knowing what to do with such sources).

This might fix #481

Issue #510, opened 2018-03-27
Rewrite overridden source lines in generated spec file


euanh:
When a source defined in the spec file is overridden by a source defined in a link, we currently search all resources for a file with the same basename as the source link in the original spec.   This means that we can't override the name of the source - the linked source must have exactly the same name as the original.

Instead, we should rewrite the source line in the generated spec file which ends up in the SRPM.   We already rewrite this file to insert patchqueue series files, so this would not be a major change.

One feature of the current behaviour is that the original source URL remains in the SRPM.   This could be an advantage when the original source points to a public repository but the link caused the source actually to be fetched from a private mirror - the generated SRPM still refers to the canonical public source.   There are various ways in which we could handle this - the simplest might be just to write in the basename for rewritten URLs and avoid the problem entirely.

PR #518, opened 2018-03-28
Rewrite all sources and patches in the spec file

mseri:
When rewriting the specs, we delete any Source and Patch entry from the spec file
and we inject:

- first the list of computed sources (link, spec)
- then the list of computed patches (link, spec)
- finally the exploded patchqueues (link).

This should solve #510 and #511

### DIFF
--- a/planex/spec.py
+++ b/planex/spec.py
@@ -373,6 +373,12 @@ class Spec(object):
         Rewrite the sources and patches in the spec file, also inserting the
         names of all patchqueue patches into it
         """
+
+        # If there is nothing to rewrite, don't rewrite!
+        if not (self._sources or self._patches or self._patchqueues):
+            return self.spectext
+
+        # If there are patches, make sure we use autosetup or autopatch
         if self._patches or self._patchqueues:
             planex.patchqueue.check_spec_supports_patchqueues(self)
 
--- a/planex/spec.py
+++ b/planex/spec.py
@@ -401,7 +401,7 @@ class Spec(object):
         # we are sorting patches and sources to avoid tripping on weird
         # centos bug like https://bugzilla.redhat.com/show_bug.cgi?id=1359084
         def sorted_by_key(dictionary):
-            """Return an iterable of key, value tuples orderedy by key"""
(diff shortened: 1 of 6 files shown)


planex/spec.py
```diff
@@ -365,18 +367,65 @@ def specpath(self):
...
+            return bool(patch.match(line) or source.match(line))
+
+        def first_index(iterable, predicate):
+            """
+            Returns the index of first value in the iterable
+            for which predicate(value) is true.
+            """
+            return next(i for i, v in enumerate(iterable) if predicate(v))
+
+        split_at = first_index(self.spectext, is_source_or_patch_line)
```
mseri: This assumes that there is at least a `Source` or `Patch` line in the spec file, but I thought that figuring out how to support specfiles not having sources or patches was not worth the effort

srowe: This will probably fail with the `vendor-drivers` spec file then. We can add a fake source if necessary.

planex/spec.py
```diff
@@ -365,18 +368,65 @@ def specpath(self):
...
+        newspec_filtered_body = (
+            line for line
+            in self.spectext[split_at+1:]
+            if not is_source_or_patch_line(line)
+        )
+
+        # we are sorting patches and sources to avoid tripping on weird
+        # centos bug like https://bugzilla.redhat.com/show_bug.cgi?id=1359084
+        def sorted_by_key(dictionary):
+            """Return an iterable of key, value tuples orderedy by key"""
```
srowe: "ordered"

planex/spec.py
mseri: The old code was not able to deal with absent sources, but it was not rewriting them either. I think we should add the fake sources, I am not sure how to decide where to put them otherwise

planex/spec.py
mseri: Aaargh, typos are killing me :dagger:

planex/spec.py
mseri: Actually, if there are no sources or patches, we could easily fix the code

mseri: Done!

PR #524, opened 2018-03-29
init: Add required definitions to Makefile template

euanh:
These fields are now needed to make a useful Makefile


planex/cmd/init.py
```diff
@@ -32,7 +32,13 @@ def create_makefile():
 
     defaults = [".PHONY: default\n",
                 "default: rpms\n",
-                "DIST?=.fc21\n",
```
mseri: Why were we using `fc21`?

Issue #208, opened 2015-04-23
planex-fetch should not write to a temporary file


euanh:
planex-fetch writes to a temporary file called `filename~`, which it moves into place after the download is complete.   This causes a race if two instances of planex-fetch are trying to download the same file.   It is also unnecessary, since we now set delete-on-error in the makefile.

euanh:
I wonder if we still need to write to a temporary file to avoid a write-after-write race.   If so, we should use NamedTemporaryFile to ensure that the filenames are unique.

PR #527, opened 2018-03-29
fetch.py: replace pycurl with requests

### LINKED ISSUE #227: Switch from urllib2 and curl to requests
http://www.python-requests.org/en/latest/

Might not be suitable for planex-fetch - we use libcurl to spool directly to disk rather than buffering tarballs in memory.

mseri:
Should fix #227

### DIFF
--- a/planex/cmd/fetch.py
+++ b/planex/cmd/fetch.py
@@ -12,7 +12,7 @@ import urlparse
 import argcomplete
 import git
 import pkg_resources
-import pycurl
+import requests
 
 from planex.link import Link
 from planex.cmd.args import common_base_parser, rpm_define_parser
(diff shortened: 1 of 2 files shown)


mseri:
Tested locally, seems to be fetching all right

mseri:
Travis is failing with:
```
E: 16, 0: No name 'Retry' in module 'requests.adapters' (no-name-in-module)
```
It must have an ancient version of requests. I see if I can update travis to pull a more recent one.

planex/cmd/fetch.py
```diff
@@ -115,35 +103,42 @@ def fetch_http(url, filename, retries):
...
-                best_effort_file_verify(tmp_filename)
-                shutil.move(tmp_filename, filename)
-                # Write an origin file for tracking.
-                with open('{0}.origin'.format(filename), 'w') as origin_file:
-                    origin_file.write('{0}\n'.format(url_string))
-                return
+    tmp_filename = filename + "~"
+    with open(tmp_filename, 'wb') as out:
+        shutil.copyfileobj(req.raw, out)
+        best_effort_file_verify(tmp_filename)
```
mseri: Note that this is not 100% equivalent to the old code, however this best effort verification will happen only if we were able to complete the download without consuming all the retries, in that case we should have received the complete file, otherwise we would have failed before

srowe: Don't these two lines need to be done once `out` is closed? i.e. one less level of indent.

mseri: Indeed, it could have not been flushed yet

planex/cmd/fetch.py
```diff
@@ -115,35 +103,42 @@ def fetch_http(url, filename, retries):
...
-                shutil.move(tmp_filename, filename)
-                # Write an origin file for tracking.
-                with open('{0}.origin'.format(filename), 'w') as origin_file:
-                    origin_file.write('{0}\n'.format(url_string))
-                return
+    tmp_filename = filename + "~"
+    with open(tmp_filename, 'wb') as out:
+        shutil.copyfileobj(req.raw, out)
+    best_effort_file_verify(tmp_filename)
+    shutil.move(tmp_filename, filename)
```
euanh: It would be great to get rid of this as it's racy: #208.   We don't need to do anything special with the output file in fetch to clean up after a failed download because we define .DELETE_ON_ERROR in Makefile.rules, so if fetch returns a nonzero exit code make will delete the output file anyway.

mseri: I am all for it, but if we write into filename directly this is still racy as both instances will try to write over the same file

mseri: How can we prevent `planex-fetch` to fetch the same file multiple times concurrently?

planex/cmd/fetch.py
```diff
@@ -121,11 +122,10 @@ def fetch_http(url, filename, retries):
...
     req.raise_for_status()
 
-    tmp_filename = filename + "~"
-    with open(tmp_filename, 'wb') as out:
+    with tempfile.NamedTemporaryFile(delete=False) as out:
         shutil.copyfileobj(req.raw, out)
-    best_effort_file_verify(tmp_filename)
-    shutil.move(tmp_filename, filename)
+    best_effort_file_verify(out.name)
+    shutil.move(out.name, filename)
```
euanh: I think you can just get rid of the temporary file altogether and write to the final destination.   If the verification fails, exit with a non-zero exit code and `make` will delete it.   If you run `planex-fetch` by hand you could end up with a 'bad' file but that could actually be an advantage because you can then see what you actually downloaded (often a redirect or error page from GitHub) without the need for another `--keeptmp` flag.

I think we added this tempfile initially because the makefile was trying to download the same file several times in parallel.  We should treat that as a bug in the makefile.

mseri: Good, let's get rid of that then

mseri:
Let's merge and see if it builds only on my machine

Issue #529, opened 2018-04-04
Add a way to remove patches and patchqueue with a pin file


mseri:
We can add and overwrite soruce/patch/patchqueues using pin files, but we cannot currently delete them. We can simply add this feature by allowing pins with empty `sourceX`, `patchX` or `patchqueueX` (or with a `delete=true` field)

Issue #515, opened 2018-03-28
Add support for v3 link schema


euanh:
Currently we are overriding the existing v2 schema parsing with the semantics of v3 in `load()`.   Add a new stanza to handle the v3 schema and return the v2 parser to strict compatibility with existing v2 links.

Issue #503, opened 2018-03-27
Refactor link file parsing logic


euanh:
Previously we loaded the contents of .lnk files into Link objects, specified in link.py.   When we added the v2 schema, we added some logic to paper over the differences between the two forms.   In this refactoring, we have changed the way we handle links - rather than consulting the link object, the Spec object has an API for adding 'resources' which handles overriding of existing resources.   This simplifies the logic in the tools and tests, so seems to be an improvement.

With this change we can get rid of the Link class entirely and load and parse the JSON in the new load() function (currently in spec.py).

mseri:
PINS version:

1. deprecated:
    - `URL`: source repository url
    - `commitish`: source repository tag/commitish
    - `patchqueue`: name appended to the generated patchqueue tarball
    - `base`: base repository url (for PQ pins, and used only by planex-clone afaik)
    - `base_commitish`: base repository commitish (for PQ pins, and used only by planex-clone afaik)
    - `sources`: folder for additional sources
    - `patches`: ??? not sure

v2.
    - `source`: unused
    - `patch`: contains `URL` and `commitish` pointing to a tarball with extra files (sources and patches)
    - `patchqueue`: contains `URL` and `commitish` pointing to a tarball with patches and a series file

v3.
    - `sourceX`: contains `URL` and `commitish` pointing to a "blob" overriding or adding  `sourceX` in the spec
    - `patchX`:  contains `URL` and `commitish` pointing to a "blob" overriding or adding `patchX` in the spec
    - `archiveX`: contains `URL` and `commitish` pointing to a tarball with extra files (used when looking for the files)
    - `patchqueueX`: contains `URL` and `commitish` pointing to a tarball with patches and a series file

mseri:
@euanh please update the above is any of the information is wrong. What was `patches` in pins v1 for?

mseri:
I have not yet deleted the Link class, I think I will postpone it for now

PR #531, opened 2018-04-05
Refactor schemaVersion to distinguish between 2 and 3, and drop support for 1

mseri:
This should solve #503 and #515.

~~This PR also introduces a way to delete a source, a patchqueue or a patch by providing an empty record in the link/pin file. I have created #529 to track that as I recall that issue being discussed at a certain point. I am happy to remove the relevant commits if we decide that this is not necessary.~~

EDIT: Removed, it will be part of a separate PR if we decide to stack pins over links

### DIFF
--- a/planex/spec.py
+++ b/planex/spec.py
@@ -291,6 +291,41 @@ class InvalidSchemaVersion(KeyError):
     pass
 
 
+def update_with_schema_version_2(spec, link):
+    """Update spec with a schemaVersion 2 link"""
+    for name, value in link.patch_sources.items():
+        idx = _parse_name(name)
+        spec.add_archive(name, value["URL"], link.path,
+                         value["patches"])
+
+    for name, value in link.patchqueue_sources.items():
+        idx = _parse_name(name)
+        spec.add_patchqueue(idx, value["URL"], link.path,
+                            value["patchqueue"])
+
+
+def update_with_schema_version_3(spec, link):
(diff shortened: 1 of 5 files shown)


planex/spec.py
```diff
@@ -319,21 +319,30 @@ def load(specpath, link=None, check_package_name=True, defines=None):
...
                                         link.path, value.get("prefix"),
                                         value.get("commitish"))
                 else:
                     spec.add_source(idx, value["URL"], link.path)
             for name, value in link.patch_sources.items():
                 idx = _parse_name(name)
-                spec.add_archive(idx, value["URL"], link.path,
-                                 value["patches"])
+                if not value:
+                    spec.rm_patch(idx)
```
euanh: Why does `rm_patch` match with `add_archive` here?   If that's what is intended could you please add a comment explaining why these two go together.

planex/spec.py
```diff
@@ -291,54 +291,71 @@ class InvalidSchemaVersion(KeyError):
...
+            spec.add_gitarchive(idx, value["URL"],
+                                link.path, value.get("prefix"),
+                                value.get("commitish"))
+        else:
+            spec.add_source(idx, value["URL"], link.path)
+
+    for name, value in link.patch_sources.items():
+        idx = _parse_name(name)
+        if not value:
+            spec.rm_patch(idx)
```
euanh: Why is `rm_patch` the opposite of `add_archive`?

Issue #505, opened 2018-03-27
Add bulk extraction for tarball-type sources


euanh:
Archive.extract_source() (defined in spec.py and inherited by Patchqueue) takes the name of a single file, creates a Tarball object, extracts the file from the tarball, then throws away the tarball.   This is very inefficient for archives which contain lots of files, such as patchqueues.

Change extract_source() to extract_sources(), which takes a list of filenames and tries to extract them all at once.  It should return a list of files which it was able to extract.   Spec.extract_sources() should then try to extract all the sources it needs from the first archive, remove those which were successfully extracted from the list and then try the second archive, and so on until it has found all the sources.

PR #532, opened 2018-04-09
Add bulk extraction for tarball-type sources

mseri:
This should solve #505

It includes a very simple caching of the archive file list (or series
list) to speedup the `__contains__` checks.


mseri:
Please let me know if you believe that this is the wrong approach

mseri:
Yes for both!

mseri:
Currently if patches and patchqueues are coming from the same archive, we are extracting the patchqueues twice. Give that the list of resources is small, I could add a commit to avoid this double extraction by pre-filtering the list of extracted files.

planex/spec.py
```diff
@@ -253,14 +274,33 @@ def extract_source(self, name, destdir):
...
 
+    def extract_sources(self, names, destdir):
+        """
+        Extract all the source [names] to [destdir].   Raises [KeyError]
+        for the first requested source that cannot be found.
+        """
+        with Tarball(self.path) as tarball:
+            for name in names:
+                target_path = os.path.normpath(os.path.join(self.prefix, name))
+                tarball.extract(target_path, destdir)
```
euanh: https://docs.python.org/2/library/tarfile.html#tarfile.TarFile.extractall might be worth a look here. You would call getmembers(), take the intersection of that and the list you are looking for, then pass that intersection to extractall().

euanh:
We need to define the search order when the same file could come from several archives.    First match or last match would make sense;  last match might be best as the working patchqueue is likely to be last and would then override any earlier patches from the package.

mseri:
Yes, I think last match as by construction patcqueues are checked last (and it's also what is happening now)

planex/spec.py
mseri: When I looked at `extractall` I was worried by

> Warning
> Never extract archives from untrusted sources without prior inspection. It is possible that files are created outside of path, e.g. members that have absolute filenames starting with "/" or filenames with two dots "..". 

And I was convinced that we where using `extractmember` for this reason here: https://github.com/xenserver/planex/blob/depend-refactor/planex/tarball.py#L56

Looking back at the code... that was not the case. So I'll go on with your suggestion

planex/cmd/makesrpm.py
```diff
@@ -127,11 +127,10 @@ def populate_working_directory(tmpdir, spec):
...
     sources = [os.path.basename(source[0]) for source in spec.sources()]
-    for source in sources:
-        try:
-            spec.extract_source(source, tmpdir)
-        except KeyError:
-            print("Could not find a source for %s" % source)
+    try:
+        spec.extract_sources(sources, tmpdir)
+    except KeyError as err:
+        print("Could not find a source for %s" % err.message)
```
srowe: `BaseException.message` is long deprecated, this should be `str(e)`

http://sayspy.blogspot.co.uk/2007/05/baseexceptionmessage-is-now-deprecated.html

planex/spec.py
```diff
@@ -553,21 +595,39 @@ def resource(self, target):
...
+            for (idx, sources) in enumerate(collection_batches)
+        ]
+
+        pending = set(sources)
+        for (resource, collection) in zip(resources, filtered_batches):
+            resource.extract_sources(collection, destdir)
+            pending -= set(collection)
+
+        if pending != set():
+            KeyError(pending)
```
srowe: `raise KeyError(...)`?

mseri: indeed! Well spotted

mseri:
Before merging I will cleanup the history further

Issue #521, opened 2018-03-29
Add ignore-autosetup to link files and re-enable autosetup check for patches


mseri:
The check for the presence of `%autosetup` or `%autopatch` has been disabled for patches by https://github.com/xenserver/planex/pull/520 because in some spec files with patches we already deal with them in a different way. This is not robust now that we can add custom additional patches via link files.

To be able to re-enable the check, a clean solution could be to add a new optional field in the v3 link file to disable the check, and disable it in the few places where it is needed.

PR #534, opened 2018-04-10
Add ignore-autosetup to link files and re-enable autosetup check for patches

mseri:
This should solve #521

### DIFF
--- a/planex/link.py
+++ b/planex/link.py
@@ -28,6 +28,11 @@ class Link(object):
         return int(self.link.get('SchemaVersion', 1))
 
     @property
+    def ignore_autosetup(self):
+        """Return the ignore autosetup value"""
+        return bool(self.link.get('IgnoreAutosetup', False))
+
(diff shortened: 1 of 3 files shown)


planex/link.py
```diff
@@ -27,6 +27,11 @@ def schema_version(self):
         # Default to v1 if not present, for legacy lnks
         return int(self.link.get('SchemaVersion', 1))
 
+    @property
+    def ignore_autosetup(self):
+        """Return the ignore autosetup value"""
+        return bool(self.link.get('IgnoreAutosetup', False))
```
srowe: Is the `bool()` wise? If the input file has

```
"IgnoreAutosetup": "False"
```

then it will silently become `True`

mseri: You are right. For some unclear reasons I was convinced that it would have evaluated the `"False"` into `False`, that is clearly false!

PR #537, opened 2018-04-13
Minor refactoring of planex-depend

mseri:
This also adds an option to display the dependency graph as a json object

### DIFF
--- a/planex/cmd/depend.py
+++ b/planex/cmd/depend.py
@@ -101,18 +101,25 @@ def package_to_rpm_map(specs):
 
 def buildrequires_for_rpm(spec, provides_to_rpm):
     """
-    Generate build dependency rules between binary RPMs
+    Generate build dependency rules between binary RPMs.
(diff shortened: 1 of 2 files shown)


planex/cmd/depend.py
```diff
@@ -101,18 +102,25 @@ def package_to_rpm_map(specs):
...
+    Generate build dependency rules between binary RPMs.
+    Returns a tuple with the rpm path and the list of
+    its buildrequires.
     """
     rpmpath = spec.binary_package_paths()[-1]
     # Package's Requires must exist for it to be installed as a
     # BuildRequire of a later package, so we make it depend on
     # Requires as well as BuildRequires to ensure they are built.
     buildreqs = (spec.buildrequires() | spec.requires()) - spec.provides()
+    buildreqrpms = []
```
srowe: This should be defined as a `set` then you don't have to exclude duplicates manually.

MarkSymsCtx: +1

Issue #540, opened 2018-04-16
Allow Patch and Patchqueues to point at git repositories


mseri:
As `SourceX` fields accept an optional `commitish` and git `URL`, Patches and Patchqueues should do the same. It will probably be enough to introduce new GitArchive and GitPatchQueue objects to deal with them

PR #541, opened 2018-04-17
Link cleanup

### LINKED ISSUE #539: Make naming more coherent in links v3
In links v3 we have some naming that seems wrong:

-  `PatchX`: does not really point to a patch but to a tar archive and, in fact, creates an `Archive` object internally
-  `PatchX` has a field `patches` and `PatchQueueX` has a field named `patchqueues`, but in both cases these are prefix for the lookup paths

To make things more coherent we should `PatchX` to `ArchiveX`, and `patches` and `patchqueues` to `prefix`.

mseri:
This PR removes leftovers of the SchemaVersion:1 parsing, improves the versioning and updates the semantics of the v3 pins.

It also raises an interesting issue with `planex-manifest`, since links v2 it has become completely useless for links and pins.

This should resolve #539 and #540

### DIFF
--- a/planex/cmd/clone.py
+++ b/planex/cmd/clone.py
@@ -10,7 +10,7 @@ from os.path import basename, dirname, join, relpath
 import subprocess
 
 import git
-from planex.link import Link
+# from planex.link import Link
 import planex.util as util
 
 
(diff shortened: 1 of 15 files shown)


planex/cmd/clone.py
```diff
@@ -116,25 +116,29 @@ def main(argv=None):
...
 
-    for pinpath in args.pins:
-        pin = Link(pinpath)
+    # for pinpath in args.pins:
+    #     pin = Link(pinpath)
 
-        if args.jenkins:
-            print('echo "Cloning %s"' % pin.url)
-            clone_jenkins(pin.url, args.repos, pin.commitish, args.credentials)
+    #     if args.jenkins:
```
srowe: Are these going to be re-added at some point?

mseri: Yes, there is an open issue to re-write `clone.py`

mseri:
I have added two commits to make planex-manifest produce some semi-usable json. I still think we should get rid of it and do something different instead

PR #542, opened 2018-04-17
Add schemaVersion for manifests

mseri:
This will help parsing the new format


tests/data/manifest/branding-xenserver.json
```diff
@@ -1,5 +1,6 @@
 {
     "sources": {}, 
+    "schemaVersion": "1",
```
mseri: This is on purpose. If you read schemaVersion 1 it means that the other fields will be empty

Issue #533, opened 2018-04-10
planex-pin should have an way to autoselect the pin path


mseri:
This feature was removed when planex-pin was merged to upstream planex and replaced by a manual `-o` option

PR #545, opened 2018-04-25
planex-pin does not know about the new pin format

### LINKED ISSUE #507: planex-pin does not know about the new pin format
The tool should be updated to emit schemaVersion v3 pin files

mseri:
This should fix #507 and #533.

The new CLI looks like the following:
```
[planex:master local]$ planex-pin --help
usage: planex-pin [-h] [--quiet] [-v] [--version] [-w | -o OUTPUT]
                  [--source-override SOURCE] [--patchqueue-override PATCHQUEUE]
                  PACKAGE
Create a .pin file for PACKAGE. Needs to run from the root of a spec
repository. Note that when URL is an ssh url to a git repository, planex will
first look for a repository with the same name cloned in the $CWD/repos
folder.
positional arguments:
  PACKAGE               package name
optional arguments:
  -h, --help            show this help message and exit
  --quiet, --warn       Only log warnings and errors
  -v, --verbose, --debug
                        Enable debug logging
  --version             show program's version number and exit
  -w, --write           Write pin file in PINS/PACKAGE.pin. It overwrites the
                        file if present.
  -o OUTPUT, --output OUTPUT
                        Path of the pinfile to write. It overwrites the file
                        if present.
  --override-source SOURCE
                        Path to a tarball or url of a git repository in the
                        form URL:commitish. When used the pin will get rid of
                        any pre-existing source, archive or patchqueue and use
                        the provided path as Source0.
  --override-patchqueue PATCHQUEUE
                        Path to a tarball or url of a git repository in the
                        form URL:commitish. When used the pin will get rid of
                        any pre-existing patchqueue and use the provided path
                        as PatchQueue0.
```

Two mutually exclusive flags allow to write on a specific path (`-o`) or on the automatically generated
`PINS/$package_name.pin` path (`-w`). The default behaviour is still to print on `stdout` only.

By default `planex-pin $pkgname` will load `$pkgname.spec` and `$pkgname.lnk` and generate a pin file with the current state of affairs from these two. This means that the new pin files are more verbose than the old one by default, but give a true snapshot of the state of things. The team pins will likely want to remove some of the autogenerated content (like the additional sources).

Two mutually exclusive flags can be used to create a fresh pin with a single `Source0` as source of truth, or a pin file where the `PatchQueue` is replaced by a custom `PatchQueue`.  These are documented in the command help.

This also re-enables `planex-clone` in `Jenkins`. Required for testing the changes.

### DIFF
--- a/planex/cmd/pin.py
+++ b/planex/cmd/pin.py
@@ -122,14 +122,13 @@ def parse_args_or_exit(argv=None):
                        help="Path of the pinfile to write. "
                             "It overwrites the file if present.")
 
-    overrs = parser.add_mutually_exclusive_group()
-    overrs.add_argument("--source-override", dest="source", default=None,
+    parser.add_argument("--override-source", dest="source", default=None,
                         help="Path to a tarball or url of a git "
                              "repository in the form ssh://GitURL#commitish. "
                              "When used the pin will get rid of any "
                              "pre-existing source, archive or patchqueue "
                              "and use the provided path as Source0.")
-    overrs.add_argument("--patchqueue-override", dest="patchqueue",
+    parser.add_argument("--override-patchqueue", dest="patchqueue",
                         default=None,
                         help="Path to a tarball or url of a git "
                              "repository in the form ssh://GitURL#commitish. "
--- a/planex/cmd/pin.py
+++ b/planex/cmd/pin.py
@@ -12,6 +12,7 @@ import sys
 
 from planex.cmd.args import common_base_parser
 from planex.link import Link
+from planex.repository import Repository
 from planex.spec import GitBlob, GitArchive, GitPatchqueue, Archive
 import planex.spec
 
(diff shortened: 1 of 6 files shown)


mseri:
Please have a second look at this. The changes here make planex-pin full-fledged and enables planex-clone to work in a restricted set of cases (including the Jenkins builds). All the remaining work should be in making `planex-clone` suitable for the developer workflow.

I suggest to merge this and release a new beta to get more thorough Jenkins testing on the main branches, and send the next batch of changes for `planex-clone` in a separate PR (as this is already growing out of proportions)

planex/cmd/pin.py
```diff
@@ -30,91 +40,125 @@ def spec_and_lnk(repo_path, package_name):
...
-        return Repository(spec_or_link.url)
-    if spec_or_link is None:
-        return None
-    else:
-        sys.exit(
-            "repository_of: got unexpected object {}".format(
-                spec_or_link))
+def repo_or_path(arg):
+    """
+    Heuristic. Parse URL:commitish into (URL, commitish) and anything else into
```
srowe: This comment and the `argparse` help text state the separator is ":" but the code uses "#".

mseri: Well spotted, thanks!

Issue #546, opened 2018-05-01
Remove planex-manifest


MarkSymsCtx:
Planex-manifest has a number of problems and some of its functionality is replaced by extract_commit in makesrpm. extract-commit should be extended to get hashes for all repositories whether they have .git-archive or not and planex-manifest should be removed.

Issue #502, opened 2018-03-27
Decide what to do with planex-manifest


euanh:
planex-manifest does not work very well and does not understand the new link file format.

Ideally, tracking what went into a build should be handled outside planex.   If we really want to keep track of the git hash of the commit which went into a package, we can use .gitattributes.

If we remove planex-manifest, we can also get rid of repository.py, most if not all of git.py and most of utils.py.

mseri:
Note: planex-manifest output is meaningless for the new link and pin files

mseri:
The current plan is the following:
- remove planex-manifest
- add metadata in the spec file to keep some relevant information (it can be disabled and removed in an easy way)
- once a better way is in place internally get rid also of that

PR #547, opened 2018-05-01
Delete planex-manifest and add metadata to the generated spec file

mseri:
This address both https://github.com/xenserver/planex/issues/546 https://github.com/xenserver/planex/issues/502 based on the discussions that we have been having in the last few days


planex/cmd/makesrpm.py
```diff
@@ -136,14 +129,20 @@ def populate_working_directory(tmpdir, spec):
...
 
     with open(newspec, "w") as out:
-        out.writelines(spec.rewrite_spec())
+        out.writelines(
+            spec.rewrite_spec(srpm_sources=srpm_sources, manifests=manifests))
 
-    if manifests:
-        add_manifest_entry(manifests, newspec)
-    else:
+    if not manifests:
```
MarkSymsCtx: We probably don't need this anymore if we're not strictly limited to reading the SHA from the .gitarchive-info.

planex/spec.py
```diff
@@ -560,10 +560,16 @@ def specpath(self):
...
+    # pylint: disable=too-many-locals
+    def rewrite_spec(self, srpm_sources=None, manifests=None):
         """
         Rewrite the sources and patches in the spec file, also inserting the
-        names of all patchqueue patches into it
+        names of all patchqueue patches into it.
+
+        If [srpm_sources] is a list of sources, it will add comments in the
+        spec file describing the origin of the source.
+        If [manifests] is true it will add `Provides: gitsha(sha) = url` using
```
MarkSymsCtx: Hopefully this is not what happens, should be `Provides: gitsha(URI) = SHA`

planex/spec.py
```diff
@@ -603,28 +609,70 @@ def sorted_by_key(dictionary):
...
+            if isinstance(pq, GitPatchqueue)
+            else "# Patchqueue: {}\n".format(pq.url)
+            for pq in patchqueues if srpm_sources is not None
+        )
+
+        if manifests is None:
+            manifests = {}
+
+        manifest_metadata = (
+            "Provides: gitsha({}) = {}\n".format(sha, url)
```
MarkSymsCtx: This is backwards. swap sha and uri

planex/spec.py
mseri: It is :P I'll update the code and fix the docstring

planex/spec.py
mseri: Oops, I'll fix it

PR #556, opened 2018-05-31
Handle str vs bytes mismatches in python3

srowe:


mseri:
Let's see how it goes. I think it may be worth adding a `_decode_if_needed` method and using it in the methods that return values from `self.spec.sourceHeader`. But if these two are the only incoherent bits I think it is ok to keep the changes to a mininum

PR #558, opened 2018-06-05
Depend refactor

srowe:
I believe this branch now as feature-parity with `master`, in order to use it in the REQ-596 pipelines I propose it should be merged and tagged as `v4.0.0`


planex/cmd/pin.py
```diff
@@ -30,91 +38,131 @@ def spec_and_lnk(repo_path, package_name):
...
-            "repository_of: got unexpected object {}".format(
-                spec_or_link))
+def repo_or_path(arg):
+    """
+    Heuristic. Parse URL#commitish into (URL, commitish) and anything else into
+    (URL, None)
+    """
+    if arg.startswith("ssh://"):
+        split = arg.split("#")
+        if len(split) > 2 or not split:
```
MarkSymsCtx: This will except if split is None, should be `if not split or len(split) > 2:` as that will short-circuit?

planex/cmd/pin.py
```diff
@@ -30,91 +38,131 @@ def spec_and_lnk(repo_path, package_name):
...
+
+    if args.patchqueue is not None:
+        url, commitish = repo_or_path(args.patchqueue)
+        pinfile["PatchQueue0"] = {"URL": url}
+        if commitish is not None:
+            pinfile["PatchQueue0"]["commitish"] = commitish
+            pinfile["PatchQueue0"]["prefix"] = resources["PatchQueue0"].prefix
+
+        # When both a PQ0 and an Archive0 are present, and point to the same
+        # repository, we assume that they are pointint to the same tarball.
```
MarkSymsCtx: typo

planex/cmd/pin.py
```diff
@@ -30,91 +38,131 @@ def spec_and_lnk(repo_path, package_name):
...
+        # This, by default, planex-pin will overwrite the Archive0 with
```
MarkSymsCtx: Thus?

planex/cmd/pin.py
mseri: Well spotted. It actually can never be `None` or an empty list at that point. It could be `if len(split) > 2:`

PR #560, opened 2018-06-05
pin: simplify repo_or_path()

srowe:
Also fix typos.

### DIFF
--- a/planex/cmd/pin.py
+++ b/planex/cmd/pin.py
@@ -51,13 +51,11 @@ def repo_or_path(arg):
     (URL, None)
     """
     if arg.startswith("ssh://"):
-        split = arg.split("#")
-        if len(split) > 2 or not split:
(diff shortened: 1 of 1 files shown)


planex/cmd/pin.py
```diff
@@ -51,13 +51,11 @@ def repo_or_path(arg):
...
-        split = arg.split("#")
-        if len(split) > 2 or not split:
-            raise ValueError(
-                "Expected URL or ssh://URL#commitish but got {}".format(arg))
-        if len(split) == 1:
-            return (arg, None)
-        return tuple(split)
+        if '#' in arg:
+            return tuple(arg.split("#", 1))
+
```
mseri: If it i not there is not an error, only if there is more than `1`.

srowe: Ah, ok. I've re-done this.

PR #565, opened 2018-09-07
spec.py: prepend source directory to sources and patches

### LINKED ISSUE #563: Duplication of source configuration between specs and links
With schema v3 it is now necessary to specify auxiliary sources in both the spec and link files. e.g.

*xapi-storage-script.spec:*
```
Source1: xapi-storage-script.service
Source2: xapi-storage-script-sysconfig
Source3: xapi-storage-script-conf.in
```

*xapi-storage-script.lnk:*
```
{
    "SchemaVersion": "3",
    "Source1": {
        "URL": "SOURCES/%{name}/xapi-storage-script.service"
    },
    "Source2": {
        "URL": "SOURCES/%{name}/xapi-storage-script-sysconfig"
    },
    "Source3x": {
        "URL": "SOURCES/%{name}/xapi-storage-script-conf.in"
    }
}
```

This is error-prone and unnecessary, the base directory will be the same for all sources.
There should be a default base directory that is added to all sources that are just a basename. This could be configured using the `.rc` file mechanism mentioned in #536

srowe:
By default add a directory to the filepath of source and patch blobs
so that local files are resolved correctly.

This default can still be overridden in a link/pin file.

Fixes #563


srowe:
I'm going use this for a full REQ-596 build before merging.

PR #568, opened 2018-09-19
CA-247468 Make makesrpm.py more self-contained

TimSmithCtx:
In order to avoid issues where "rpmbuild -bs" as invoked by makesrpm.py
is trying to mess with the RPM _topdir directories as used by other
parts of the process at the same time, separate that invocation of
"rpmbuild -bs" into its own separate _topdir (it was already partway
there).

While doing this, separate all the "--define" arguments into individual
makefile variables and supply them only where they are really needed. We
are supplying "--define _topdir" to the mock wrapper because it does
extre repository creation work which relies on expanding "%_rpmdir", but
then we have explicit core to avoid passing it on to mock because it
would break things. We *were* passing "%_sourcedir" on to mock and
getting away with it because it was ignoring us on a rebuild operation.

This reduces the risk that new arguments added to a "used everywhere"
variable will break things.


planex/cmd/makesrpm.py
```diff
@@ -240,11 +248,14 @@ def main(argv=None):
 
     args = parse_args_or_exit(argv)
     tmpdir = tempfile.mkdtemp(prefix="px-srpm-")
+    for subdir in [ 'BUILD', 'BUILDROOT','RPMS','SOURCES','SPECS','SRPMS' ]:
+        os.mkdir(os.path.join(tmpdir, subdir))
```
srowe: Are these directories guaranteed not to exist? If not use `util.makedirs()`

planex/cmd/makesrpm.py
```diff
@@ -240,11 +248,14 @@ def main(argv=None):
...
 
     try:
         specfile = populate_working_directory(tmpdir, args.spec, args.link,
                                               args.sources, args.patchdata)
-        sys.exit(rpmbuild(args, tmpdir, specfile))
+        if rpmbuild(args, tmpdir, specfile) == 0:
```
srowe: If `rpmbuild()` returns non-zero will `makesrpm`'s exit code now reflect that?

planex/cmd/makesrpm.py
TimSmithCtx: They are being made in the tmpdir created by tempfile.mkdtemp() in the line above, so unless tempfile.mkdtemp() has a rather serious problem, they are guaranteed not to exist.

planex/cmd/makesrpm.py
TimSmithCtx: Point. Have updated.

tests/specs/Makefile
```diff
@@ -5,7 +5,7 @@ PINSDIR ?= PINS
 SPECS = $(wildcard $(PINSDIR)/*.spec SPECS/*.spec)
 LINKS = $(wildcard $(PINSDIR)/*.pin SPECS/*.lnk)
 MOCK_CONFIGDIR ?= mock
-RPM_EXTRA_DEFINES=--define "_sourcedir %_topdir/SOURCES/%name"
```
srowe: This line also appears in `xenserver-specs/Makefile`, will it need changing there too?

TimSmithCtx: RPM_EXTRA_DEFINES has been superseded by RPM_SOURCEDIR_DEFINE, but now planex/Makefile.rules will set RPM_SOURCEDIR_DEFINE if it isn't set already.

So assuming there isn't some unexpected jiggery-pokery which bypasses that, RPM_SOURCEDIR_DEFINE will only need setting in xenserver-specs/Makefile if it should be different.

I think RPM_EXTRA_DEFINES *should* be removed in xenserver-specs/Makefile for clarity, but has become harmlessly ignored.

srowe: I think `RPM_EXTRA_DEFINES` has been used in the past to pass random defines into `rpmbuild`. They look to have all gone now, we should remove it to prevent people being misled.

TimSmithCtx: Not before merging and deploying this change, however, as everything will break :-)

srowe:
Is there a similar changeset for `master`?

PR #569, opened 2018-09-19
CP-28152 replace planex jenkins file with json

samuelconnolly:
planex-clone output a .groovy syntax file onto the command line and now
instead outputs a json file to be consumed by the build object in the
new pipelines

### DIFF
--- a/planex/cmd/clone.py
+++ b/planex/cmd/clone.py
@@ -4,7 +4,6 @@ planex-clone: Checkout sources referred to by a pin file
 from __future__ import print_function
 
 import errno
-from string import Template
 import argparse
 from os import symlink
 from os.path import basename, dirname, join, relpath, splitext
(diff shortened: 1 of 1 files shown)


planex/cmd/clone.py
```diff
@@ -70,13 +70,13 @@ def repo_name(url):
 """)
 
 
-def clone_jenkins(url, destination, commitish, credentials):
-    """Print Jenkinsfile fragment to clone repository"""
-    destination = join(destination, repo_name(url))
-    print(CHECKOUT_TEMPLATE.substitute(url=url,
```
srowe: You can delete the definition of `CHECKOUT_TEMPLATE` above and also remove the `--credentials` argument.

MarkSymsCtx:
Nice side effect that the planex code doesn't need to understand how to handle credentials now, win :)

PR #570, opened 2018-09-19
Simplify dev workflow

srowe:
Various improvements to make the usual developer workflows not require unnecessary arguments or resources.


srowe:
I've pinned every package using the current link definitions but they need to be updated to remove the unnecessary `Archive0`. I'll then build the REQ-596 `master` branch and `team/ring0/master` to check the pinning & cloning in Jenkins.

srowe:
`master` builds ok. I think I'll merge and then test pinning with Sam's clone changes.

PR #571, opened 2018-09-20
CP-28152 add output option to planex-clone

samuelconnolly:
Used to state where to put clone_json if alternate location would be
preferred, also defaults to . so it writes config file to pwd

### DIFF
--- a/planex/cmd/clone.py
+++ b/planex/cmd/clone.py
@@ -5,7 +5,7 @@ from __future__ import print_function
 
 import errno
 import argparse
-from os import symlink
+from os import symlink, getcwd
 from os.path import basename, dirname, join, relpath, splitext
 import subprocess
 import sys
@@ -30,6 +30,9 @@ def parse_args_or_exit(argv=None):
                       help="Print Jenkinsfile fragment")
     mode.add_argument("--clone", action="store_true",
                       help="Clone all the clonable repositories (default)")
+    parser.add_argument("--output", "-o",
+                        default=join(getcwd(), "clone_sources.json"),
+                        help="Choose output name for clone sources json file")
     parser.add_argument(
         "-r", "--repos", metavar="DIR", default="repos",
         help='Local path to the repositories')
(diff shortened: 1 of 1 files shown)


planex/cmd/clone.py
```diff
@@ -42,6 +42,8 @@ def parse_args_or_exit(argv=None):
                            "link the sources and the patchqueue and use "
                            "guilt to apply all the additional patches "
                            "[experimental]")
+    parser.add_argument("--output", "-o", default=".",
```
srowe: I think an output filename rather than a directory would be clearer.

samuelconnolly: I was going for a fixed filename since I used a fixed one in the build scripts admittedly.

srowe: As the caller consumes the file it better for that to define the entire filename.

planex/cmd/clone.py
```diff
@@ -165,7 +167,7 @@ def clone_all(args, pin):
                  "name but different commitish is not supported.")
 
     if args.jenkins:
-        clone_jenkins(gathered)
+        clone_jenkins(args, gathered)
```
srowe: Only `output` is needed to pass in `args.output`

planex/cmd/clone.py
```diff
@@ -42,6 +42,8 @@ def parse_args_or_exit(argv=None):
                            "link the sources and the patchqueue and use "
                            "guilt to apply all the additional patches "
                            "[experimental]")
+    parser.add_argument("--output", "-o", default="./clone_sources",
```
srowe: Can the default have a `.json` suffix, and the leading `./` is unnecessary.

planex/cmd/clone.py
```diff
@@ -55,12 +57,12 @@ def repo_name(url):
...
 
 
-def clone_jenkins(gathered):
+def clone_jenkins(filename, gathered):
     """Print json file containing repositories to clone"""
     json_dict = {}
     for url, commitish in gathered:
         json_dict[repo_name(url)] = {'URL': url, 'commitish': commitish}
-    with open("clone_sources.json", "w+") as clone_sources:
+    with open(filename, "w+") as clone_sources:
```
srowe: Just `"w"` should be sufficient.

PR #576, opened 2018-10-05
CP-28152 fix multiple pins not being appended

samuelconnolly:
multiple pin files were causing the json output to be overwritten, check
for existing file


planex/cmd/clone.py
```diff
@@ -62,9 +62,12 @@ def repo_name(url):
 
 def clone_jenkins_json(filename, gathered):
     """Print json file containing repositories to clone"""
-    json_dict = {}
+    json_dict = dict()
     for url, commitish in gathered:
         json_dict[repo_name(url)] = {'URL': url, 'commitish': commitish}
+    if exists(filename):
```
srowe: If you read the file before the `for` loop you will ensure any value from this pin replaces what was already there.

planex/cmd/clone.py
```diff
@@ -62,9 +62,12 @@ def repo_name(url):
 
 def clone_jenkins_json(filename, gathered):
     """Print json file containing repositories to clone"""
-    json_dict = {}
+    json_dict = dict()
```
srowe: This change should be a no-op.

PR #586, opened 2018-11-30
Add "show" and "unpin" options to planex-pin

TimSmithCtx:
Quality-of-life improvements for planex-pin. To show the current state
of a pin, use "planex-pin --show <package>", and to remove a pin, use
"planex-pin --unpin <package>".

This removes one need for the user to know the implementation of pins.

### DIFF
--- a/planex/cmd/pin.py
+++ b/planex/cmd/pin.py
@@ -8,6 +8,7 @@ import argparse
 import json
 import os
 import sys
+import errno
 
 from planex.blobs import Archive
 from planex.cmd.args import common_base_parser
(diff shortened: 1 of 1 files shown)


planex/cmd/pin.py
```diff
@@ -139,15 +147,32 @@ def main(argv=None):
...
+            if e.errno not in (errno.ENOENT,):
+                raise
+            print("Package '{}' is not pinned".format(package_name))
+            pass
+
+    if args.unpin:
+        try:
+            os.remove(default_output)
+        except:
+            pass
```
mseri: I think it would be better to fail ungracefully if the removal goes wrong

TimSmithCtx: For any other reason than ENOENT, maybe. Unpinning a thing which is already not pinned should be a no-op.

srowe: Why not test for the existence of the file first? This isn't a command where we need to worry much about concurrency.

TimSmithCtx: It was easier to just catch the error, given that I have some code doing just that right above it to borrow.

srowe:
Can you add a SoB?

PR #588, opened 2018-12-07
Correct error path on RequestException

TimSmithCtx:
The length of the args array when a RequestException occurs is not
known. If sticking to descriptors inherited from BaseException, it would
be better to use 'message' to avoid an exception while handling the
exception.

### DIFF
--- a/planex/cmd/fetch.py
+++ b/planex/cmd/fetch.py
@@ -161,7 +161,7 @@ def fetch_url(url, source, retries):
     except requests.RequestException as exn:
         # Download failed
         sys.exit("%s: Failed to fetch %s: %s" %
-                 (sys.argv[0], urlunparse(url), exn.args[1]))
+                 (sys.argv[0], urlunparse(url), str(exn)))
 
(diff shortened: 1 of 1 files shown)


planex/cmd/fetch.py
```diff
@@ -161,7 +161,7 @@ def fetch_url(url, source, retries):
     except requests.RequestException as exn:
         # Download failed
         sys.exit("%s: Failed to fetch %s: %s" %
-                 (sys.argv[0], urlunparse(url), exn.args[1]))
+                 (sys.argv[0], urlunparse(url), exn.message))
```
srowe: I thought `BaseException.message` had been deprecated

https://www.python.org/dev/peps/pep-0352/#retracted-ideas

TimSmithCtx: exn.args[0] would also work. That seems to have had its deprecation retracted (my head is spinning)

srowe: What's wrong with `str(exn)`?

PR #590, opened 2019-01-17
Make it easier for users to install the package

psafont:
Now pip can take care of installing the dependencies too, making it easier to install.

I've deleted `requirements.txt` since it didn't have pinned versions of the dependencies. The same could be done with `test-requirements.txt` by adding those requirements as extra dependencies. See https://github.com/EMBL-EBI-TSI/aap-client-python/blob/master/setup.py#L76 for an example.

Further ahead, tox could be used to do the testing, and the installation to virtualenv and symlinking could be reduced to a single commands with pipx after the code is ported to python 3 and system-site-packages switch is added to it.
(`pipx --system-site-packages --spec git+https://www.github.com/xenserver/planex.git planex`)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-virtualenv: 
+virtualenv:
   system_site_packages: true
 sudo: required
 services: docker
@@ -9,9 +9,9 @@ addons:
     - yum
     - python-rpm
 install:
-  - pip install --upgrade -r requirements.txt
+  - pip install --upgrade .
   - pip install -r test-requirements.txt
-script: 
+script:
   - pycodestyle --show-source --show-pep8 setup.py planex tests
   - pylint setup.py planex tests
   - nosetests --verbose --with-coverage --cover-inclusive --cover-package=planex
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,8 @@ RUN yum -y install yum-plugin-ovl \
 # Install basic prerequisites for building planex
 RUN yum -y install \
   epel-release \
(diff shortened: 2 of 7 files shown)


psafont:
Build is failing because dependencies are installed through setuptools now and the setuptools used (https://pypi.org/project/setuptools/0.9.8/) is incompatible with python packages built as universal wheels.

In this case the package that triggers it first is pathlib2-2.3.3

psafont:
I checked and the oldest setuptools that works is [30.6.0](https://pypi.org/project/setuptools/36.2.0/)

I know how to circumvent the distro's packaging and install a newer version using a python-bootstrapped pip, this works for the centos container.

However this does not work for the planex.spec, how can we force that to be the minimal version? (the prepackaged version of setuptools in centos 7 is from 2013 (0.9.8))

psafont:
Note that pip is actually generating wheels and then installing them on the system:

```
Building wheels for collected packages: planex
  Building wheel for planex (PEP 517): started
  Building wheel for planex (PEP 517): finished with status 'done'
  Stored in directory: /tmp/pip-ephem-wheel-cache-v7eQ70/wheels/02/21/cd/d8edd61b6567f44024abf43feed04ea8aa1fa2d9c41dbdf0e1
Successfully built planex
Installing collected packages: argparse, scandir, pathlib2, planex
  Running setup.py install for scandir: started
    Running setup.py install for scandir: finished with status 'done'
Successfully installed argparse-1.4.0 pathlib2-2.3.3 planex-4.1.4 scandir-1.9.0
```

PR #594, opened 2019-01-29
Prepare for 4.1.3

samuelconnolly:
Fix planex-clone issue where we couldn't clone certain pins due to path
issues, as well as files being treated as archives. Also stop it from complaining about previously existing clones when a re-clone is attempted.

### DIFF
--- a/planex/cmd/clone.py
+++ b/planex/cmd/clone.py
@@ -149,6 +149,9 @@ def clone(url, destination, commitish):
 
     if not destination.parent.exists():
         destination.parent.mkdir(parents=True)
+    if destination.exists():
+        # Take a fresh clone if we are re-cloning
+        shutil.rmtree(str(destination))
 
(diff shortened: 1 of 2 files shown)


planex/cmd/clone.py
```diff
@@ -216,9 +219,10 @@ def extract(url_str, destination):
         fetch_url(url, archive_path, 5)
 
     # extract the archive
-    tara = tarfile.open(archive_path)
-    tara.extractall(str(destination))
-    tara.close()
+    if tarfile.is_tarfile(archive_path):
+        tara = tarfile.open(archive_path)
```
MarkSymsCtx:
```
with tarfile.open(archive_path) as tara:
    tara.extractall(str(destination)
```
??

samuelconnolly: oh I was completely on autopilot and didn't even take that in. Yeah there's a tons of open and closes from someone unfamiliar with context handlers.

PR #608, opened 2019-05-24
CA-314309 Allow per-host configuration of repo parsers

TimSmithCtx:
Make the repository manager read the server-type property from the
hosts' section of the config file.

Since the bbtoken code uses ~/.planex as a place to keep its cache, add
~/.planex/planexrc as a search path for the config file, ahead of
~/.planexrc.

Signed-off-by: Tim Smith <tim.smith@citrix.com>

### DIFF
--- a/planex/config.py
+++ b/planex/config.py
@@ -10,6 +10,7 @@ class Configuration(object):
     """Represents a set of configuration options"""
     # pylint: disable=R0903
     searchPath = ('/etc/planexrc',
+                  os.path.expanduser('~/.planex/planexrc'),
                   os.path.expanduser('~/.planexrc'),
                   '.planexrc')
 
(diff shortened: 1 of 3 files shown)


planex/config.py
```diff
@@ -10,6 +10,7 @@ class Configuration(object):
     """Represents a set of configuration options"""
     # pylint: disable=R0903
     searchPath = ('/etc/planexrc',
```
samuelconnolly: I feel like this should be a self variable.

samuelconnolly:
Unit tests need updating to deal with new import but functionally looks fine.

TimSmithCtx:
> Unit tests need updating to deal with new import but functionally looks fine.

Disabling the unit tests for now. Will raise task to work out how to re-enable them.

PR #611, opened 2019-06-19
CA-318284 Wildcard expansion fails internally

samuelconnolly:
*.pin will get passed down to planex-clone if nothing is expanded on the
terminal, just check for wildcards in filenames and exit quietly.


samuelconnolly:
Ah test regression introduced by Tim's change previously. Fun.

samuelconnolly:
> Ah test regression introduced by Tim's change previously. Fun.

.name is a member of MockRepo now, since this was added in the cmp_repos function. Probably should expand that test more too...

PR #612, opened 2021-03-31
CA-353005: fix source issues with PINs

MarkSymsCtx:
* Archive the git repo in expected format
 * Archive to the filename expected by the specfile


MarkSymsCtx:
The travis CI should ideally be replaced by Github actions.

TimSmithCtx (approved):

planex/blobs.py
```diff
@@ -118,7 +118,8 @@ class GitBlob(Blob):
     """
 
     # pylint: disable=too-many-arguments
-    def __init__(self, spec, url, defined_by, prefix, commitish):
+    def __init__(self, spec, url, defined_by, spec_path, prefix, commitish):
+        self.spec_path = spec_path
```
edwintorok, line 122: I assume that `spec_path` here means the path of the Source URL from the `.spec`, not the path of the `.spec` itself, and indeed when called below it gets a source path passed in. For clarity this could be called source_path or spec_source_path (to avoid confusion with the `spec` variable).

planex/blobs.py
```diff
@@ -147,15 +148,7 @@ def commitish(self):
...
-
-        # RPM only looks at the basename part of the Source URL - the
-        # part after the rightmost /.   We must match this behaviour.
-        #
-        # Examples:
-        #    ssh://git@www.example.com/foo/bar.git -> bar.tar.gz
-
-        (name, _) = os.path.splitext(os.path.basename(self.url))
-        return os.path.join("%_sourcedir", "{}.tar.gz".format(name))
+        return os.path.join("%_sourcedir", self.spec_path)
```
edwintorok, line 151: This removes the guessing of the filename, good!

edwintorok (approved):

planex/blobs.py
MarkSymsCtx: It's the spec version of this path pre-override.

